PR #69, opened 2015-11-25
Create and Display News Items

itsWill:
Screenshot of what it looks like: 
screen shot 2015-11-24 at 9 08 41 pm

## Note: Doesn't have support for translations

Tackles it #5 
@cmm246 thanks for the help!

@nicholasjhenry  for review


sophiedeziel:
There are some conflicts with master. Can you resolve them?
Also, can you display news items on the home page as well?

itsWill:
Added  validations to the NewsItem. We only display published news items on the news page and the front page.  @sophiedeziel  for review

app/controllers/application_controller.rb
```diff
@@ -2,11 +2,13 @@ class ApplicationController < ActionController::Base
...
-  before_filter :get_next_meetup
+  before_filter :get_next_meetup, :get_events
 
   def get_next_meetup
     @next_meetup = Meetup.next
   end
 
-
+  def get_events
+    @events = Event.published
```
sophiedeziel: I think that variable should be assigned in `HomeController` because it is used in `app/views/home/index.html.erb`. The other variable `@next_meetup` in the method up there is something that was previously used by the application layout so that why it's there.

itsWill: It's actually used in the [sidebar](https://github.com/montrealrb/Montreal.rb/blob/master/app/views/layouts/_sidebar.html.erb#L2-L17) as well to display upcoming events. Since the sidebar is part of the application layout, I moved it to the `application_controller`

sophiedeziel: Oh! I did not see that! Alright you made the right decision!!!

PR #70, opened 2015-11-25
Adds job posting to administrate

anonyo:


sophiedeziel:
There are merge conflicts with master. I guess you are already working on solving them?

spec/factories/jobs.rb
```diff
@@ -0,0 +1,9 @@
+FactoryGirl.define do
+  factory :job do
+    state "MyString"
+title "MyString"
+description "MyText"
+organization_id 1
+  end
+
+end
```
sophiedeziel: Can we correct indentation here?

anonyo:
@sophiedeziel this is good to go!

app/dashboards/dashboard_manifest.rb
```diff
@@ -11,6 +11,7 @@ class DashboardManifest
   # to work correctly.
   DASHBOARDS = [
     :locations,
+    :jobs
     :events,
   ]
```
nicholasjhenry: Hey @anonyo, I assume this is working, but this doesn't look like a valid array, missing a comma after `jobs`.

app/views/fields/enum_field/_form.html.erb
```diff
@@ -0,0 +1,2 @@
+<%= f.label field.attribute %>
+<%= f.select field.attribute, f.object.class.public_send(field.attribute).options %>
```
nicholasjhenry: @anonyo very slick :boom:

app/dashboards/dashboard_manifest.rb
anonyo: Hey @nicholasjhenry, my bad, that definitely wouldn't work, fixed.

app/views/fields/enum_field/_form.html.erb
anonyo: Thanks to you and @benichu

benichu: :+1:

cawel:
@kiramclean and I bumped into the same problem when implementing the "pages". Indeed, to this day, [Administrate]((https://github.com/thoughtbot/administrate)) does not support the `enum` data type by default. We got some inspiration from an [Administrate's issue](https://github.com/thoughtbot/administrate/issues/150), but did not go all the way through. We're glad you guys did :raised_hands:

@anonyo: are you going to submit a PR to the [Administrate project](https://github.com/thoughtbot/administrate)? Otherwise I will :)

anonyo:
@cawel that's a great idea, I'll work on that!

sophiedeziel:
@anonyo Yes!! I highly recommend it! :smile_cat:

benichu:
@cawel @anonyo don't forget that the custom `Administrate` enum field we created for the workshop is generic BUT specific to the gem `enumerize` not the default rails enum feature... It might be a problem if you try and submit it to the `Administrate` project team...

cawel:
Yeah, I thought about that too. Thanks for the heads up @benichu.

Things to keep in mind when using Rails' `enum`:

* the data type for the enum field has to be an `Integer`, as opposed to a `String` (otherwise it does not work)
* retrieving the possible values for the enum field is done using the [pluralized form of the enum field](http://api.rubyonrails.org/v4.2.3/classes/ActiveRecord/Enum.html), e.g. `Page.states`, as opposed to calling [Enumerize's `options` method](https://github.com/brainspec/enumerize/blob/b81a5593f56d0e35967e53f25ae84f0c27bef718/lib/enumerize/attribute.rb#L57-L75).

PR #74, opened 2015-11-25
Manage and display members (#8)

Coaxial:
Add a Member model + tests and add it to the admin panel

[Review on Reviewable](https://reviewable.io/reviews/montrealrb/montreal.rb/74)


spec/factories/members.rb
```diff
@@ -0,0 +1,11 @@
...
+  factory :member do
+    name "MyString"
+email "MyString"
+picture ""
+twitter_handle "MyString"
+github_handle "MyString"
+biography "MyText"
+  end
+
+end
```
sophiedeziel: Ca we correct indentation here please?

Coaxial:
cc @GalkaS

nicholasjhenry:
@Coaxial Would love to merge. Are you up for rebasing with master?

Coaxial:
I'm getting weird conflicts, it looks like unresolved things were committed:

```
diff --cc app/dashboards/event_dashboard.rb
index 48b312c,715bbe6..0000000
--- a/app/dashboards/event_dashboard.rb
+++ b/app/dashboards/event_dashboard.rb
@@@ -10,28 -10,12 +10,34 @@@ class EventDashboard < Administrate::Ba
    ATTRIBUTE_TYPES = {
      location: Field::BelongsTo,
      id: Field::Number,
 +<<<<<<< HEAD
 +<<<<<<< HEAD
++<<<<<<< HEAD
++=======
++>>>>>>> 359248fe323be0350abb0ebbe192da3c120e431d
      starts_at: Field::DateTime,
      created_at: Field::DateTime,
      updated_at: Field::DateTime,
      title: Field::String,
      introduction: Field::Text,
      conclusion: Field::Text
++<<<<<<< HEAD
 +=======
 +    type: Field::String,
 +    starts_at: Field::DateTime,
 +    created_at: Field::DateTime,
 +    updated_at: Field::DateTime,
 +>>>>>>> #14 : Add events to administrate
 +=======
 +    starts_at: Field::DateTime,
 +    created_at: Field::DateTime,
 +    updated_at: Field::DateTime,
 +    title: Field::String,
 +    introduction: Field::Text,
 +    conclusion: Field::Text
 +>>>>>>> #14 : Add title, introduction, conclusion.
++=======
++>>>>>>> 359248fe323be0350abb0ebbe192da3c120e431d
    }
  
    # COLLECTION_ATTRIBUTES
@@@ -40,20 -24,10 +46,26 @@@
    # By default, it's limited to four items to reduce clutter on index pages.
    # Feel free to add, remove, or rearrange items.
    COLLECTION_ATTRIBUTES = [
 +<<<<<<< HEAD
++<<<<<<< HEAD
++=======
++>>>>>>> 359248fe323be0350abb0ebbe192da3c120e431d
      :id,
      :title,
      :starts_at,
      :location,
++<<<<<<< HEAD
 +=======
 +    :location,
 +    :id,
 +    :starts_at,
 +<<<<<<< HEAD
 +>>>>>>> #14 : Add events to administrate
 +=======
 +    :title,
 +>>>>>>> #14 : Add title, introduction, conclusion.
++=======
++>>>>>>> 359248fe323be0350abb0ebbe192da3c120e431d
    ]
  
    # SHOW_PAGE_ATTRIBUTES
@@@ -64,34 -38,17 +76,46 @@@
    # an array of attributes that will be displayed
    # on the model's form (`new` and `edit`) pages.
    FORM_ATTRIBUTES = [
 +<<<<<<< HEAD
++<<<<<<< HEAD
++=======
++>>>>>>> 359248fe323be0350abb0ebbe192da3c120e431d
      :title,
      :introduction,
      :conclusion,
      :starts_at,
      :location,
++<<<<<<< HEAD
 +=======
 +    :location,
 +    :starts_at,
 +<<<<<<< HEAD
 +>>>>>>> #14 : Add events to administrate
 +=======
 +    :title,
 +    :introduction,
 +    :conclusion
 +>>>>>>> #14 : Add title, introduction, conclusion.
++=======
++>>>>>>> 359248fe323be0350abb0ebbe192da3c120e431d
    ]
  
    # Overwrite this method to customize how events are displayed
    # across all pages of the admin dashboard.
    #
 +<<<<<<< HEAD
++<<<<<<< HEAD
 +  def display_resource(event)
 +     event.title
 +  end
 +=======
 +  # def display_resource(event)
 +  #   "Event ##{event.id}"
 +  # end
 +>>>>>>> #14 : Add events to administrate
++=======
+   def display_resource(event)
+      event.title
+   end
++>>>>>>> 359248fe323be0350abb0ebbe192da3c120e431d
  end
```

nicholasjhenry:
@Coaxial I was able to rebase this on my machine. Let's me know if you need a hand, happy to help. I didn't see any unresolved commits. Please let me know if you see any.

benichu:
@Coaxial Would you like us to rebase that branch before merging?
Thanks for your help :)

nicholasjhenry:
@Coaxial and I will work on this tomorrow. 

cc @benichu

nicholasjhenry:
This is ready for review if anybody is up for it.

cawel:
Here are a few comments in disguise of a code review:

### More important:

* since the `email` and `name` attributes are required to be present at the application level (in the model), I would also enforce that at the database level with non-null db constraints (well, at least for the `email`).
* I would add a db unique index for that `email` column
* I see that a `User` `has_one :member`, and that a `Member` `has_one :user`. However, neither model has a foreign key. Fishy ! :fish:

### Less important:

* I think it's more idiomatic to write:

```ruby
expect(member).to_not be_valid
```

instead of:

```ruby
expect(member.valid?).to be false
```

* except for the uniqueness validation, it is not necessary to _persist_ the model in order to check for the validation errors. You can do:

```ruby
member = Member.new(email: nil)
expect(member).to_not be_valid
```

Terser, and also tests will run faster :rocket:

nicholasjhenry:
To expand on @cawel idiomatic comments:

```
expect(member).to_not be_valid
```

can also be written as

```
expect(member).to be_invalid
```

Positive expectations/assertions are often a little easier to read.

cawel:
Better indeed.

Coaxial:
Thank you for the feedback, much appreciated. We'll work on it tonight with @galkas

cawel:
@nicholasjhenry should probably jump in before you refactor tonight, as he had a clear idea in which db table to put that foreign key.

app/models/member.rb
```diff
@@ -0,0 +1,8 @@
+class Member < ActiveRecord::Base
+  has_many :events
+  has_many :organizations
+  has_one :user
```
GalkaS: Should we write a test for this association?

nicholasjhenry: @GalkaS I don't test the presence of associations specifically -- higher level tests will verify if this association exists -- i.e the app will blow up pretty quickly when doing a request spec for example.

nicholasjhenry:
> @nicholasjhenry should probably jump in before you refactor tonight, as he seemed to have (during the workshop) a clear idea in which db table to put that foreign key (either in 'users' table or 'members' table).

Sorry I couldn't address this earlier. The foreign key should really be placed on the member. i.e. a `member has_one user`. The reason for this is that we want to have our dependencies point towards the user class, not the user class having dependencies on other classes in our system. This avoids the user class becoming a [god object](https://en.wikipedia.org/wiki/God_object).

I understand that you placed it on the user as the foreign key in Administrate doesn't support `belongs_to` relationships out of the box. I have sometime to look at this tomorrow so I will do some investigation on what the solution to this is.

Thank you for taking the time to make this amendment, and again sorry I didn't respond earlier.

cawel:
>  The foreign key should really be placed on the member. i.e. a `member has_one user`.

Note that based on his explanation, @nicholasjhenry probably meant to write: a `member *belongs_to* user`.

nicholasjhenry:
@cawel :+1:

nicholasjhenry:
OK, just shows how confused I was :scream_cat: 

The current version of Administrate we have does not support `has_one` relationships. So moving the foreign_key to `members.user_id` will work as that will be a `belongs_to` relationship. Let me know if you're up for changing this. Thank you!

Thanks to @cawel for keeping my associations speak honest.

Coaxial:
Got it, we'll make the change this weekend :godmode:

db/migrate/20151206020220_remove_foreign_key_from_users_and_add_foreign_key_to_members.rb
```diff
@@ -0,0 +1,9 @@
+class RemoveForeignKeyFromUsersAndAddForeignKeyToMembers < ActiveRecord::Migration
```
karimmtarek: Isn't it be better if you create one migration for each table change?

Coaxial: Yeah, we weren't sure so we put them in the same migration because they are related. We can split in two, no problem

karimmtarek: @Coaxial :+1:

app/models/event.rb
```diff
@@ -17,6 +17,7 @@ class Event < ActiveRecord::Base
 
   validates :title, presence: true
   validates :starts_at, presence: true
+  has_many :members
```
sophiedeziel: I tried to create a member from the dashboard. Once the user created, I got an error because the key is missing. 

This will be used for RSVPs? I suggest to remove that `:has_many` line and add it back once we implement the RSVP feature

screen shot 2015-12-09 at 10 28 39 pm

app/models/member.rb
```diff
@@ -0,0 +1,8 @@
+class Member < ActiveRecord::Base
+  has_many :events
```
sophiedeziel: see https://github.com/montrealrb/Montreal.rb/pull/74/files#r47185636

app/dashboards/member_dashboard.rb
```diff
@@ -0,0 +1,62 @@
...
+class MemberDashboard < Administrate::BaseDashboard
+  # ATTRIBUTE_TYPES
+  # a hash that describes the type of each of the model's fields.
+  #
+  # Each different type represents an Administrate::Field object,
+  # which determines how the attribute is displayed
+  # on pages throughout the dashboard.
+  ATTRIBUTE_TYPES = {
+    events: Field::HasMany,
+    organizations: Field::HasMany,
```
sophiedeziel: After playing around to understand administrate and that bug: https://github.com/montrealrb/Montreal.rb/pull/74/files#r47185636

I found out that removing all references in the dashboard to events and organizations for now fixes it.

app/dashboards/member_dashboard.rb
```diff
@@ -0,0 +1,62 @@
...
+
+  # COLLECTION_ATTRIBUTES
+  # an array of attributes that will be displayed on the model's index page.
+  #
+  # By default, it's limited to four items to reduce clutter on index pages.
+  # Feel free to add, remove, or rearrange items.
+  COLLECTION_ATTRIBUTES = [
+    :events,
+    :organizations,
+    :user,
```
sophiedeziel: After removing `:events` and `:organization`, some admin pages could not be rendered because the user dashboard is not in the manifest and it tries to create a link to the user dashboard. For now, we should replace `:user` for `:name`.

Coaxial: We're a bit confused. We should replace `:user` with `:name` in the manifest, correct? We're not familiar with the code written for the `User` model, should we just remove `:votes` from the manifest but keep the column in the db?

karimmtarek: @Coaxial, please: 
- Rebase with master and fix the conflicts.
- Don't replace `:user` with `:name` that was a hack to get it to work, do the right thing ;)
- Get familar with the user model :)
- It's better if you remove the `user_dashboard` and generate a new one based on the last changes like so: `rails g administrate:dashboard user`
- Let me know if you need a hand to finish this

sophiedeziel:
@Coaxial Do you mind if I rebase on master or do you want to do it yourself?

Coaxial:
Did I do this wrong? I did `git fetch && git rebase origin/master` and had a ton of conflicts to resolve. Now there are a ton of commits that aren't ours in the PR...

sophiedeziel:
Yes, the conflicts are normal. About the commits, I think it's because of the chronological order of the ones that were already merges to master. Should be ok.

There are still some amendment required before merging to master (our reviews and houds warning). Do you want to do them?

sophiedeziel:
Seems like the conflicts were not resolved by the rebase...

Coaxial:
@sophiedeziel I dont know how to resolve the conflicts without pulling in the hundred and something commits from master

sophiedeziel:
I'm giving you a dirty shortcut that works:
```
git checkout master
git pull --rebase
git checkout feature/members
git merge master
```
You will have conflicts. Resolve them, make all the tests pass and then
```
git commit
```

sophiedeziel:
@Coaxial are you on slack right now?

sophiedeziel:
I fully tested the feature on my machine and I fixed stuff in a way that doesn't break anything if deployed. 

However, you will notice in `app/dashboards/member_dashboard.rb` that I commented out the user fields. The reason is because Administrate will raise an error on `create`, `update` and `show` because it needs a user dashboard and controller. The dashboard is there but the controller is missing (the route too). 

@Coaxial, can you create that part so members can be fully managed and associated to users and to manage the user record too?

Issue #14, opened 2014-09-29
Manage and display events


vivrass:
As an organizer, I want to be able to be able to create a monthly event

    Choose a meet-up place
    Choose 2 talks
    Add an introduction + conclusion (french + english)

- [x] manage events in the admin
- [x] display events
- [x] render the show view
- [x] add missing attributes to the index view
- [x] link title on the index view

nmoller:
I'm on it.

nicholasjhenry:
Admin completed for this, thanks! We just need to display events on the site.

nicholasjhenry:
Some outstanding tasks on this:

- [ ] render the show view
- [ ] add missing attributes to the index view
- [ ] link title on the index view

benichu:
I'll take care of those tasks (on the NewsItem model) in my migration branch #42 as this is helpful for data cleanup

PR #78, opened 2015-11-25
#14 Display events.

nmoller:


Gemfile
```diff
@@ -37,6 +37,8 @@ gem 'rails_12factor', group: :production
 gem 'globalize'
 gem 'globalize-accessors'
 
+gem 'friendly_id'
```
sophiedeziel: ❤️

nicholasjhenry: Just a heads up there is a lighter version of this that Rails provides which has a couple of benefits:
- don't need a third party gem
- don't need to modify query code, i.e. `find`

Example:

```ruby
class Event
   to_param :title
end
```

I haven't used this with Globalize before but assume it should work.

sophiedeziel: This is a good opportunity to try the behavior of both solutions when using Globalize. If they both work without creating a mess of config, we'll have to make a decision based on:

- We should use the same strategy for all models, so what is the effort to implement it?
- How do we manage uniqueness? (that one is very easy but both solutions have completely different solutions. with to_param, we should raise a validation error and change the value manually by resubmitting the form. With friendly_id, we can define the rules to generate a slug without requiring any other action from the user)

nmoller: Hi Nicholas and Sophie,
Nich, the solution you propose was the first Martin suggested. We werent able to make it works so we used friendly_id.
I'll look at the alternative way to see if I can make it work.
Hey, I want to thank you for the organisation, it was awesome.

sophiedeziel: I personally love friendly_id because of the way we can define how to generate slugs and manage uniqueness. If you need some help to try both solutions with Globalize, I can help. Ping me on slack and I'll answer when I am available.

nicholasjhenry: Just a heads-up, you don't need to worry about uniqueness with Rails `to_param` as it does include the ID in the param. It just works. Obviously not as cool as FriendlyID, but a lot simpler.

Anyway, just a suggestion - I always prefer to use less gems if possible -- other opinions may differ :smile:. 

Another consideration is FriendlyID quickly pollutes the code base with:

` Event.friendly.find(param)`

But with a code base this size it probably doesn't matter.

nicholasjhenry: > Nich, the solution you propose was the first Martin suggested. We werent able to make it works so we used friendly_id.

Ah interesting, I wonder want prevented it from working.

> I'll look at the alternative way to see if I can make it work.

Sounds good.

> Hey, I want to thank you for the organisation, it was awesome.

You're most welcome! Thank you for coming out. We forming a great a team here and with @sophiedeziel building on @vivrass's awesome efforts over the last 4 1/2 years.

nicholasjhenry: OK, I've confirmed that Rails `to_param` works with Globalize and Adminstrate. As mentioned previously, I like Rails `to_param` as it's extremely light weight.

https://github.com/montrealrb/Montreal.rb/commit/21dfe9633a2dbc9da4b2eae91496e4a12dc9b2e2

But @sophiedeziel has the last call on the stack so I will defer to her. I'll shut up now :smile_cat:

cawel: Can't be more "light weight" than this, as it's already built-in in the framework ;)

Here are my 2 cents, so that we can reach @sophiedeziel's dollar some day! Light weight is certainly good from a design point of view :+1: But I want to emphasize a small detail. Due to the implementation of `to_param`, when it comes to routing, anything after the dash (`-`) in the url will be ignored. For example, if you have a page model with `id` = 1, then `/pages/1-ignored-garbage-characters` would be a perfectly routable url, as the id (1 in this case) is what counts. From Google's point of view, that means it's theoretically possible to create pages with duplicate content, which is bad SEO. Also, having the `id` in the url is also not optimal.

Bottom line: my guess is that SEO is not that important for that project, so I would also be in favor of using `to_param`, at least until the day this project needs a tad more sophisticated url slugs. 

That said, it's now my turn to shut up :smile_cat:

mathieugagne: If SEO is a concern, we can add meta tags about canonical urls to reach the same content and add a sitemap. Google shouldn't penalize us for it. Wordpress has many different urls for a single post and works similarly. Same actions are needed there as well.

2 more cents :p

On Nov 27, 2015, at 02:12, Martin Carel <notifications@github.com<mailto:notifications@github.com>> wrote:


In Gemfile<https://github.com/montrealrb/Montreal.rb/pull/78#discussion_r46021357>:

> @@ -37,6 +37,8 @@ gem 'rails_12factor', group: :production
>  gem 'globalize'
>  gem 'globalize-accessors'
>
> +gem 'friendly_id'


Can't be more "light weight" than this, as it's already built-in in the framework ;)

Here are my 2 cents, so that we can reach @sophiedeziel<https://github.com/sophiedeziel>'s dollar some day! Light weight is certainly good from a design point of view [:+1:]  But I want to emphasize a small detail. Due to the implementation of to_param, when it comes to routing, anything after the dash (-) in the url will be ignored. For example, if you have a page model with id = 1, then /pages/1-ignored-garbage-characters would be a perfectly routable url, as the id (1 in this case) is what counts. From Google's point of view, that means it's theoretically possible to create pages with duplicate content, which is bad SEO. Also, having the id in the url is also not optimal.

Bottom line: my guess is that SEO is not that important for that project, so I would also be in favor of using to_param, at least until the day this project needs a tad more sophisticated url slugs.

That said, it's now my turn to shut up [:smile_cat:]

-

nmoller:
You have to integrate it to the events_controller (try to implement the show and index actions), it was at that level that we had some difficulties. 
We had issues at that level... Thanks for the input, I'll try that...

Gemfile
nicholasjhenry: Thanks for jumping @cawel -- not 100% sure I understand those points, but happy to discuss on another channel rather than extending this long thread.

nicholasjhenry:
Thanks @nmoller -- I'll give that a go. More for my own interest than anything else -- good to know what is compatible and what's not.

PR #77, opened 2015-11-25
Add basic pages scaffold

kiramclean:
Depends on enum type existing in administrate (depends on PR #70).

PR #79, opened 2015-11-26
Pages cleanup

cawel:
New stuff in this PR:

* Removed useless code + useless files from Rails' scaffold
* Added basic validations in the `Page` model
* Ensured a requested page is 'published' (otherwise it should return a 404)
* Added controller and model specs

Note that this PR depends on PR #77 (which is the reason why this PR has a merge commit).

Tagging @kiramclean in case she still wants to be kept in the loop.


spec/models/page_spec.rb
```diff
@@ -0,0 +1,26 @@
+require 'rails_helper'
+
+RSpec.describe Location, :type => :model do
```
nicholasjhenry: @cawel This should be `Page` not `Location`. :lollipop:

Gemfile
```diff
@@ -27,6 +27,7 @@ gem 'turbolinks'
 
 # View
 gem 'simple_form'
+gem 'enumerize'
```
nicholasjhenry: `enumerize` is already included in the Gemfile

https://github.com/montrealrb/Montreal.rb/blob/master/Gemfile#L23

db/migrate/20151125004633_create_pages.rb
```diff
@@ -0,0 +1,11 @@
+class CreatePages < ActiveRecord::Migration
```
nicholasjhenry: @cawel Would you be up to make title and body "globalized"? If not happy to accept the PR as is.

https://github.com/globalize/globalize#creating-translation-tables

Gemfile
cawel: Fixed the offending commit using a rebase interactive.

db/migrate/20151125004633_create_pages.rb
cawel: Fixed the offending commit using a rebase interactive. The `db/schema.rb` should be up to date.

spec/models/page_spec.rb
cawel: Fixed the offending commit using a rebase interactive.

cawel:
@nicholasjhenry All fixes done, sir! :rocket:

benichu:
@nicholasjhenry what's the status on that PR? Can we merge it?

Gemfile
nicholasjhenry: @cawel Unfortunately `enumerize` is still included in this pull-request. In master it's located here:

https://github.com/montrealrb/Montreal.rb/blob/master/Gemfile#L23

nicholasjhenry:
This is good to merge once the following issue has been resolved:

https://github.com/montrealrb/Montreal.rb/pull/79/files#r46279753

Gemfile
cawel: @nicholasjhenry My understanding of the issue with the `enumerize` gem was that it was _duplicated_ in the Gemfile. So I fixed it a few days ago ([here is the relevant commit](https://github.com/cawel/Montreal.rb/commit/ac85398e4157ca02d14a0c4ad528c7130fdbf722])), changing the Gemfile to include it only once.

Both my `migrations-cleanup` branch and my `pages-cleanup` branch have that gem only once in the Gemfile. I'm not sure about what you expect from me. Is there something I misunderstood?

nicholasjhenry: @cawel I'm understand your confusion. I thought this had been rebased with `master` recently. I'm not sure the last time you rebased, but when we merged PR the `enumerize` gem will appear twice in the Gemfile. Please view below:

1. https://github.com/cawel/Montreal.rb/blob/pages-cleanup/Gemfile#L29
2. https://github.com/montrealrb/Montreal.rb/blob/master/Gemfile#L23

My advice is to rebase again, please. Then the Gemfile in your branch should not include an entry for `enumerize`.

cawel: I rebased both of my branches on top of `master`. And that Gemfile looks the same to me :)

nicholasjhenry: So apparently I'm an idiot (or just my multitasking skills are failing me). As @cawel pointed out the `enumerize` gem was moved, not added -- I missed the deletion for some reason.

cawel: No worries @nicholasjhenry ! 

Look 'ma, it's merged! :tada:

PR #82, opened 2015-11-27
Migrations cleanup

cawel:
* Fixed a few minor things (e.g. typo + missing translations table drop statements).
* Made sure it is possible to revert _all_ migrations all the way to the very beginning.


db/migrate/20151124183140_remove_talks.rb
```diff
@@ -1,5 +1,6 @@
 class RemoveTalks < ActiveRecord::Migration
   def change
-    drop_table :talks
+    drop_table :talks do
```
nicholasjhenry: @cawel I'm unsure why an empty block is needed here. What am I missing?

benichu: probably a typo, `drop_table` does not execute blocks.

```ruby
def drop_table(table_name, options = {})
  execute "DROP TABLE #{quote_table_name(table_name)}"
end
```

Source: http://apidock.com/rails/v4.2.1/ActiveRecord/ConnectionAdapters/SchemaStatements/drop_table

benichu: @cawel can you remove those empty blocks and we can merge that PR right away. Thanks a lot!

cawel: Not a typo. When trying to rollback a migration which had a simple `drop_table` statement, I had this error:

```
To avoid mistakes, drop_table is only reversible if given options or a block (can be empty).
```

Consequently, I decided to add an empty block (choosing one of the 2 options) and that made the said migration reversible. And that was the goal of that commit, because I think there is value in having all migrations reversible, whenever possible.

So @benichu, do you still want me to remove those empty blocks?

sophiedeziel: Wow, does that mean that the rails docs are outdated?

cawel: No, it's here:

http://apidock.com/rails/ActiveRecord/Migration/CommandRecorder/invert_drop_table

benichu: @cawel I guess we can always learn something :)
Thanks for investigating.
I am AFK but if anyone can merge, please go ahead...

cawel: Sure thing :) I was also surprised to bump into that. :open_mouth:

Note that this empty block is only needed when `drop_table` is used within a `change` method in the migration. Within a `down` method, the simple `drop_table` statement (w/o a block) would do.

I have just added a comment in the code as to why we need an empty block, in each of the 2 relevant migrations. PR updated using `push -f`.

benichu: I hate that kind of magic, when: what you see is what you __don't__ get
But maybe it's because I am old and grumpy :)

cawel: Totally. I guess we're both old and grumpy then :boom:

sophiedeziel: With the project I am working on right now, I had to do a drop_table. In the block, I specified the actual fields in the block. So the migration is reversible and what you see is what you get. Example: 

```
drop_table :users do |t|
  t.integer :id, index: true  
  t.string :username
  t.string :password
end
```

Of course, this is not the table I dropped.

nicholasjhenry: That makes sense to me, otherwise you don't really have a reversible migration.

benichu: It does make sense, But i guess that instead of a having a method's behavior hijacked because of the context where it is executed, I would rather have used `up` and `down` in place of `change` like in the old days!
But, as I said, I am just grumpy :older_man:

It wouldn't happen with Phoenix :trollface:

nicholasjhenry:
@benichu I agree on that.

I guess what I was pointing to was that the block passed to `drop_table` has a purpose and that we should either leave it out -- so it is a irreversible, or add the required schema to make it truly irreversible -- not to just avoid the exception. A mis-guiding exception message for sure.

cawel:
@nicholasjhenry Very valid points. I have just added a commit to that PR to make those migrations Truly Reversible :tm:

cawel:
I also retested to make sure it's possible to 'rake db:rollback` all the way down to the very beginning, and it still works.

benichu:
@nicholasjhenry can we merge this PR too?

nicholasjhenry:
This PR is based on the branch for PR #79. It can be merged as well once this fix is made:

@cawel Unfortunately `enumerize` is still included in this pull-request. In master it's located:

https://github.com/montrealrb/Montreal.rb/pull/79/files#r46279753

PR #85, opened 2015-11-28
Issue #14 : using to_param for event

nmoller:


app/models/event.rb
```diff
@@ -12,6 +12,8 @@
 class Event < ActiveRecord::Base
   translates :title, :introduction, :conclusion
   belongs_to :location
+  # suggestion from Nicholas
+  to_param :title
```
benichu: @nicholasjhenry If we are going to use `to_param` to generate pretty urls for pages representing objects in the application, it might be useful to define a common standard (mixin?)

Also, in the case of that particular implementation, I think that it would be something more like:
```ruby
def to_param
  "#{id} #{title}".parameterize
end
```
Allowing us to have urls like these: `/events/1-my-event-title` where Rails would automatically be able to load the `Event` object like so `event = Event.find("1-my-event-title")`

Finally, because I know you LOVE gems, would there be any added value for us to use a specialized pretty urls library like https://github.com/norman/friendly_id ?

nicholasjhenry: @benichu are you trolling me? 😈 The class method does exactly what you described above.

benichu: I totally missed the `.to_param` class method, very badly tested the behavior in my console :disappointed: 
BUT I was definitely trolling you about `friendly_id` because, you know, I have too much time on my hands :smiling_imp:

benichu: As a reference for my future self: http://apidock.com/rails/ActiveRecord/Integration/ClassMethods/to_param

nicholasjhenry:
Thank you @nmoller. Just a couple of things. Let me know if you would like to add them, or otherwise I'm happy to accept the PR as is:

- [ ] Render the events index with title, started_at, description
- [ ] Render the event show with title, started_at, description

nmoller:
Hi Nicholas,
Thanks for your work. I apologize for not adding the rspecs to make it
pass at 100%...
If you allows me that time to do it, I'm working on learning to test my
code, so you can accept as it is.




>>> Nicholas Henry <notifications@github.com> 01/12/15 9:27 >>>
Thank you @nmoller. Just a couple of things. Let me know if you would
like to add them, or otherwise I'm happy to accept the PR as is:
  
*   Render the events index with title, started_at, description 
*   Render the event show with title, started_at, description 
 —
Reply to this email directly or view it on GitHub.
     
NaN.

nicholasjhenry:
I'm merging this in as, I believe @cawel work on #89 resolves the missing render requirements.

Issue #17, opened 2014-09-29
Register as member a via Github 


vivrass:
Allow users to register with Github

PR #86, opened 2015-11-30
Implement omniauth-github

em77:
Implemented omniauth-github to resolve issue #17 and allow users to sign up with GitHub. I tested this with my own GitHub key and secret and it seems to be working but I've returned those values to the default `GITHUB_KEY` and `GITHUB_SECRET` which need to be replaced with those from Montreal.rb's account in `config/initializers/devise.rb`.

[Review on Reviewable](https://reviewable.io/reviews/montrealrb/montreal.rb/86)


em77:
Good call. Looks like it's still up to the server admin to set the environment variables. Let me know if there's any other changes I should make.

benichu:
@em77 Thanks for that PR, we will look into it asap, taking care of setting up the appropriate GITHUB ENV variables on our heroku server.

cc @sophiedeziel

sophiedeziel:
We should also have in the Readme file, in the setup part, a list of the environment variables needed for some features.

nicholasjhenry:
See my recommendation to use app.json to help with some of this:

https://github.com/montrealrb/Montreal.rb/issues/104#issuecomment-161998990

em77:
Added some setup instructions for the environment variables to the README, let me know what you think.

sophiedeziel:
Reviewed 10 of 21 files at r1, 2 of 4 files at r3, 1 of 4 files at r5.
Review status: 13 of 22 files reviewed at latest revision, 13 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/montrealrb/montreal.rb/86)*

PR #87, opened 2015-12-01
refactored news_item to remove #published?

itsWill:
Used a lambda in the validation felt that it was more concise and slightly cleaner than using a method. Though a could
``` ruby
def published?
  state.published?
end
```
also be used.

@cawel thanks for the tip!

### DIFF
--- a/app/models/news_item.rb
+++ b/app/models/news_item.rb
@@ -1,15 +1,10 @@
 class NewsItem < ActiveRecord::Base
   extend Enumerize
   enumerize :state, in: [:draft, :archived, :published]
-  validates :published_at, presence: true, if: :published?
+  validates :published_at, presence: true, if: lambda{ state.published? } 
   validates_presence_of :title, :state
 
   def self.published
     where(state: "published").order(:published_at)
-  end
-
-  def published?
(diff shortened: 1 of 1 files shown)


cawel:
:+1: for the lambda.

Bonus: you could also turn that class method into a [scope](http://guides.rubyonrails.org/active_record_querying.html#scopes).

itsWill:
:+1: neat idea.

cawel:
Oh yeah. Always nice to see clean code :)

app/models/news_item.rb
```diff
@@ -1,15 +1,7 @@
 class NewsItem < ActiveRecord::Base
   extend Enumerize
   enumerize :state, in: [:draft, :archived, :published]
-  validates :published_at, presence: true, if: :published?
+  scope :published, -> { where(state: :published).order(:published_at) }
+  validates :published_at, presence: true, if: lambda{ state.published? }
```
nicholasjhenry: @itsWill Thanks for making this change. One suggestion would be to use consistent lambda syntax. In the scope, the "stabby lambda" `->` (preferred), and in the validation the keyword `lambda` is used.

PR #89, opened 2015-12-02
Markdown support (phase 1)

cawel:
New stuff:

* Add markdown support for those models: `Event`, `NewsItem`, `Organization`, `Page`.

Note 1: when the 'show' route was missing, in an attempt to minimize confusion between features (different people working simultaneously on different features), I created the 'show' template _only_ when the corresponding (non-admin) controller was present in the codebase. Hence, I left out the `Talk` and `Job` models.

Note 2: the `filter` attribute has been ignored, as was requested by @benichu .

Code reviews are welcome of course, even tough it's pretty basic at this point.

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,7 +1,2 @@
 module ApplicationHelper
-  @@renderer = Redcarpet::Markdown.new(Redcarpet::Render::HTML, autolink: true, tables: true)
-
-  def render_markdown_as_html(markup)
-    raw @@renderer.render(markup)
-  end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,2 +1,7 @@
 module ApplicationHelper
+  @@renderer = Redcarpet::Markdown.new(Redcarpet::Render::HTML, autolink: true, tables: true)
+
+  def render_markdown_as_html(markup)
+    raw @@renderer.render(markup)
+  end
 end
--- a/Gemfile
+++ b/Gemfile
@@ -27,6 +27,7 @@ gem 'turbolinks'
 # View
 gem 'simple_form'
 gem 'enumerize'
+gem 'redcarpet'
 
 # Server
 gem 'unicorn'
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -229,6 +229,7 @@ GEM
     rb-fsevent (0.9.6)
     rb-inotify (0.9.5)
       ffi (>= 0.5.0)
+    redcarpet (3.3.3)
     responders (2.1.0)
       railties (>= 4.2.0, < 5)
     rest-client (1.8.0)
@@ -338,6 +339,7 @@ DEPENDENCIES
   pry
   rails (~> 4.2.2)
   rails_12factor
+  redcarpet
   rspec-rails
   sass-rails
   simple_form
--- a/app/controllers/events_controller.rb
+++ b/app/controllers/events_controller.rb
@@ -1,17 +1,11 @@
 class EventsController < ApplicationController
-  before_action :load_event, only: [:show]
 
   def index
     @events = Event.published
   end
 
   def show
-
+    @event = Event.find(params[:id])
   end
-
-  private
-    def load_event
-      @event = Event.find(params[:id])
-    end
 
 end
--- a/app/controllers/news_items_controller.rb
+++ b/app/controllers/news_items_controller.rb
@@ -2,4 +2,8 @@ class NewsItemsController < ApplicationController
   def index
     @news_items = NewsItem.published.to_a
   end
+
+  def show
+    @news_item = NewsItem.find(params[:id])
+  end
 end
--- a/app/controllers/news_items_controller.rb
+++ b/app/controllers/news_items_controller.rb
@@ -4,6 +4,6 @@ class NewsItemsController < ApplicationController
   end
 
   def show
-    @news_item = NewsItem.find(params[:id])
+    @news_item = NewsItem.published.find(params[:id])
   end
 end
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -2,4 +2,8 @@ class OrganizationsController < ApplicationController
   def index
     @organizations = Organization.all
   end
+
+  def show
+    @organization = Organization.find(params[:id])
+  end
 end
--- a/app/helpers/text_helper.rb
+++ b/app/helpers/text_helper.rb
@@ -1,0 +1,7 @@
+module TextHelper
+  @@renderer = Redcarpet::Markdown.new(Redcarpet::Render::HTML, autolink: true, tables: true)
+
+  def render_markdown_as_html(markup)
+    raw(@@renderer.render(markup))
+  end
+end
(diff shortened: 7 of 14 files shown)


app/helpers/application_helper.rb
```diff
@@ -1,2 +1,7 @@
 module ApplicationHelper
+  @@renderer = Redcarpet::Markdown.new(Redcarpet::Render::HTML, autolink: true, tables: true)
+
+  def render_markdown_as_html markup
+    raw @@renderer.render markup
```
nicholasjhenry: @cawel I see you're a fan of the Seattle style use of parentheses, but based on the style guide we've adopted:

> Omit parentheses around parameters for methods that are part of an internal DSL (e.g. Rake, Rails, RSpec), methods that have "keyword" status in Ruby (e.g. attr_reader, puts) and attribute access methods. Use parentheses around the arguments of all other method invocations.

https://github.com/bbatsov/ruby-style-guide#no-dsl-parens

cawel: @nicholasjhenry Good catch. And I'm sorry I overlooked that detail. Fixed using a `push -f` on that PR.

Suggestion: maybe the project could benefit from using a static code analyzer like [rubocop](https://github.com/bbatsov/rubocop)?

benichu: @nicholasjhenry had mentionned houndci.
It might do the job... Let me create a ticket for this.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,7 @@
 module ApplicationHelper
+  @@renderer = Redcarpet::Markdown.new(Redcarpet::Render::HTML, autolink: true, tables: true)
```
itsWill: Whats the advantage of having the class variable here as opposed to just defining it in the helper method below as a regular variable?

cawel: Good question! 

Class variables can be useful for initialization data. In this case, I'm defining a class variable on a module. By doing that, I ensure that the "html renderer" will be instantiated only once per rails app, as that class variable (`@@renderer`) of the `ApplicationHelper` module will be *shared* among all classes which will include that module (namely, all controllers and all views). In effect, that class variable of that `ApplicationHelper` module will also become a class variable of any class including that `ApplicationHelper` module.

Proof:

```ruby
module MyModule
  @@var = 0
  class << self
    def var
      @@var
    end
  end

  def increment_var
    @@var += 1
  end
end

class ClassA
  include MyModule
end

class ClassB
  include MyModule
end

puts MyModule.class_variables.inspect
puts MyModule.var

ClassA.new.increment_var
puts MyModule.var

ClassB.new.increment_var
puts MyModule.var
```

will output:

```ruby
[:@@var]
0
1
2
```

And to answer your question more directly, if the renderer would be defined in the instance method of the module, then a new renderer would be instantiated each time that method would be called. So defining the renderer as a class variable is a matter of both performance and memory efficiency.

Note that using class variables for class state (as opposed to initialization data) can be a bad idea. [This article](http://www.sitepoint.com/class-variables-a-ruby-gotcha/) gives an example of this, if you're curious.

itsWill: Fantastic answer, thanks :100:! I was somewhat confused on how class variables worked

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,7 @@
 module ApplicationHelper
```
nicholasjhenry: Since `ApplicationHelper` can be a bit of a dumping ground would it make sense to extract this out to a `TextHelper` module?

cawel:
@nicholasjhenry My understanding is that the main feature of the [markup gem](https://github.com/github/markup) is to support a bunch of different markups. Which is nice, but in our case, I don't think it's necessary. 

I think there is value in supporting _a single markup_, i.e. the very popular markdown. Wouldn't be confusing if some views would be in [textile](http://www.textism.com/tools/textile/), others in [creole](http://wikicreole.org/), and yet others in [markdown](http://daringfireball.net/projects/markdown/)? Then to modify a view you'd need to know the syntax of that particular markup the creator of the view used. On the other hand, <assumption>all programmers know/should know markdown.</assumption>

It is certainly not a strong opinion. Happy to being challenged.

app/helpers/application_helper.rb
cawel: Done. And I took the opportunity to rebase the `markdown-support` branch on top of `master`.

nicholasjhenry:
@cawel: 

> I think there is value in supporting a single markup, i.e. the very popular markdown. Wouldn't be confusing if some views would be in textile, others in creole, and yet others in markdown?

Yes, it would be very confusing. That wasn't my intention to support multiple markup languages -- they  are only supported if you install the appropriate gems.

I was considering from a code highlighting point of view.

> Syntax highlighting is performed on code blocks. See github/linguist for more information about syntax highlighting.

But probably best to add this if it's required in the future.

nicholasjhenry:
@cawel The other consideration was that many folks first encounter with Markdown is through GitHub and I'm assuming that this gem supports GitHub's markdown flavour:

https://help.github.com/articles/github-flavored-markdown/

cawel:
@nicholasjhenry Well, I tried to integrate that [markup gem](https://github.com/github/markup), but I realized that it's expecting a file – it really feels like a command line tool. Looking at the gem's code, it looks like we could circumvent that by passing in a dummy filename with a 'markdown' extension, but it's still a smell (i.e. misaligning with the gem's usage intention).

I tested it and it works. Here is what the code would look like:

```ruby
require 'github/markup'
module TextHelper

  def render_markdown_as_html(markup)
    raw(GitHub::Markup.render('dummy-filename.markdown', markup))
  end
end
```

I'm not a fan. You?

nicholasjhenry:
@cawel Thanks for checking this out. I wonder if this would work:

`GitHub::Markup::Markdown.new.render("*hello world*")`

in any case, let's not worry about this for now. Would rather get this merged. Do you want to rebase this with master?

Thanks!

cawel:
@nicholasjhenry Unfortunately it does not. The first argument to that `render` method should be a filename. Bottom line: let's keep using `Redcarpet` directly for now.

And I have just rebased the `markdown-support` branch on top of the `master` branch.

nicholasjhenry:
@cawel 

> Unfortunately it does not. The first argument to that render method should be a filename. 

Thanks for look at that.

>  Bottom line: let's keep using Redcarpet directly for now.

I believe I suggested that in my previous comment.

app/helpers/application_helper.rb
cawel: @itsWill FYI: I bumped into [that old blog post](http://snippets.aktagon.com/snippets/299-class-variables-vs-class-instance-variables-in-ruby) today. It's a simple code snippet, but I think it drives home the distinction between *class variables* and *class instance variables*, which is related to our previous discussion.

Note that Hound CI made me modify that Helper code to use a *class instance variable* instead of a *class variable*. Not a bad recommendation, and nor does it undermine the aforementioned performance and memory efficiency benefits.

benichu:
:+1:

nicholasjhenry:
@cawel thanks for your work on this and humouring me with my feeble gem suggestions :-)

cawel:
@nicholasjhenry Sure thing. And don't you dare stop firing away your suggestions! :)

app/helpers/application_helper.rb
benichu: @cawel I just added rubocop for running local checks, it's in `master`

Issue #58, opened 2015-11-24
Implement Authentication


nicholasjhenry:
The administration tool should require an authenticated user.

karimmtarek:
me and @em77 @marina101 we'll work on that one

karimmtarek:
I'm not sure if [that pull]( https://github.com/montrealrb/Montreal.rb/pull/75) request resolves this issue or there is more work needs to be done? please advice

karimmtarek:
@nicholasjhenry if there are still some work needs to be done to close this I can work on that, please let me know.

cawel:
@karimmtarek PR #75 indeed allows a user to sign in/out, and even includes nice capybara specs :) 

But as it stands now, nothing prevents a user *not signed in* to actually browse the admin interface (the pages under "/admin"). And I believe this is what that issue was about.

karimmtarek:
@cawel thanks for the feedback I'll work on that, should be done soon.

nicholasjhenry:
- [ ] authorize users to access the back-office

PR #95, opened 2015-12-02
Add Auth to Admin Area

karimmtarek:
- Added a hack for admin user (to be fixed based on advice), for this one I wasn't sure how to handle:
  * Should I add a new admin(boolean) column to the user table?
  * Should I add a a new type(string) column to the user table?  
- Redirect to admin area if user is admin
- Added logic for admin authentication
- Added a link to admin area if admin user is signed-in
- Specs will be fixed and new ones will be added when I fix the admin user situation :)

### DIFF
--- a/app/controllers/admin/application_controller.rb
+++ b/app/controllers/admin/application_controller.rb
@@ -8,7 +8,8 @@ class Admin::ApplicationController < Administrate::ApplicationController
   before_filter :authenticate_admin
 
   def authenticate_admin
-    # TODO Add authentication logic here.
+    return handle_unauthenticated_user unless current_user.present?
+    handle_unauthorized_user unless current_user.admin?
   end
 
   # Override this value to specify the number of elements to display at a time
@@ -16,4 +17,24 @@ class Admin::ApplicationController < Administrate::ApplicationController
   # def records_per_page
   #   params[:per_page] || 20
   # end
+
+  private
+
+  def handle_unauthenticated_user
+    unauthenticated_user_flash
+    redirect_to new_user_session_path
+  end
+
+  def handle_unauthorized_user
+    unauthorized_user_flash
+    redirect_to root_path
+  end
+
+  def unauthenticated_user_flash
+    flash[:alert] = "Be a good guest now, you should not go there!"
+  end
+
+  def unauthorized_user_flash
+    flash[:alert] = "Sorry, you just don't have access rights!"
+  end
 end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,11 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   before_filter :get_next_meetup, :get_events
 
+  def after_sign_in_path_for(_resource)
+    return admin_root_path if current_user.admin?
+    root_path
+  end
+
   def get_next_meetup
     @next_meetup = Meetup.next
   end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -115,18 +115,19 @@ ActiveRecord::Schema.define(version: 20151204172708) do
   end
 
   create_table "users", force: :cascade do |t|
-    t.string   "email",                  default: "", null: false
-    t.string   "encrypted_password",     default: "", null: false
+    t.string   "email",                  default: "",    null: false
+    t.string   "encrypted_password",     default: "",    null: false
     t.string   "reset_password_token"
     t.datetime "reset_password_sent_at"
     t.datetime "remember_created_at"
-    t.integer  "sign_in_count",          default: 0,  null: false
+    t.integer  "sign_in_count",          default: 0,     null: false
     t.datetime "current_sign_in_at"
     t.datetime "last_sign_in_at"
     t.inet     "current_sign_in_ip"
     t.inet     "last_sign_in_ip"
     t.datetime "created_at"
     t.datetime "updated_at"
+    t.boolean  "admin",                  default: false, null: false
   end
 
   add_index "users", ["email"], name: "index_users_on_email", unique: true, using: :btree
--- a/spec/features/authentication_spec.rb
+++ b/spec/features/authentication_spec.rb
@@ -1,30 +1,69 @@
 require 'rails_helper'
 
-describe "the signin process", :type => :feature do
-  let(:user) { User.create(:email => 'user@example.com', :password => 'password') }
+describe "Authentication" do
+  let(:user) { create(:user) }
+  let(:admin) { create(:admin) }
 
-  scenario "Signs me in" do
-    visit root_path
-    click_link 'Login'
-    fill_in 'Email', with: user.email
-    fill_in 'Password', with: user.password
-    click_button 'Log in'
-    expect(current_path).to eq(root_path)
-    expect(page).to have_content('Signed in successfully')
+  context "Guest user" do
+    scenario "Can sign up" do
+      visit new_user_registration_path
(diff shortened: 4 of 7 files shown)


cawel:
I would be in favor of a simple 'admin' boolean attribute on the `User` model, as I don't think we need more than 2 roles in that project (non-admin user and admin user) in the short term and even in the mid term. In which case you could remove the `admin?` method you created in the `User` model (rails generates such method automatically for boolean attributes). And in your migration, make sure that the 'admin' attribute is not nullable and defaults to `false` (i.e. a regular user by default):

```
add_column :users, :admin, :boolean, null: false, default: false
```

But I want that decision to be confirmed by @nicholasjhenry , as he is the Data Model Boss (:tm:) on that project!

nicholasjhenry:
This PR is related to #58

karimmtarek:
@nicholasjhenry unless you advise otherwise I'll move forward with @cawel suggestion, please let me know, thanks.

nicholasjhenry:
I have a preference for leaving booleans to operating states (i.e. on/off) and using explicit enums for things like roles and life-cycle states (more than one state change).

But as @cawel points out we can change this later -- so happy to go with a boolean.

db/schema.rb
```diff
@@ -20,8 +20,8 @@
   create_table "event_translations", force: :cascade do |t|
     t.integer  "event_id",     null: false
     t.string   "locale",       null: false
-    t.datetime "created_at",   null: false
-    t.datetime "updated_at",   null: false
+    t.datetime "created_at"
+    t.datetime "updated_at"
```
nicholasjhenry: @karimmtarek Can you please re-run you migrations from the beginning? Timestamps have a null constraint and it looks as if your schema was generated from an older version. Thanks!

nicholasjhenry:
Thanks for putting this together @karimmtarek. I'll do a proper review of this tomorrow.

db/schema.rb
karimmtarek: @nicholasjhenry  sure, I'll do that and push it back

app/controllers/admin/application_controller.rb
```diff
@@ -8,7 +8,8 @@ class Admin::ApplicationController < Administrate::ApplicationController
   before_filter :authenticate_admin
 
   def authenticate_admin
-    # TODO Add authentication logic here.
+    return no_access_guest if current_user.nil?
+    no_access_user if !current_user.admin?
```
nicholasjhenry: @karimmtarek This one would be a nice style violation to fix:

```ruby
no_access_user if !current_user.admin?
```

should be

```ruby
no_access_user unless current_user.admin?
```

app/controllers/application_controller.rb
```diff
@@ -4,6 +4,11 @@ class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception
   before_filter :get_next_meetup, :get_events
 
+  def after_sign_in_path_for(resource)
```
nicholasjhenry: @karimmtarek also this one :smile_cat:

app/controllers/admin/application_controller.rb
```diff
@@ -8,7 +8,8 @@ class Admin::ApplicationController < Administrate::ApplicationController
   before_filter :authenticate_admin
 
   def authenticate_admin
-    # TODO Add authentication logic here.
+    return no_access_guest if current_user.nil?
```
nicholasjhenry: Since the line below will be using `unless` -- I tend to make multiple guard clauses symmetrical - so you could change this to:

```
return no_access_guest unless current_user.present?
```

app/controllers/admin/application_controller.rb
```diff
@@ -16,4 +17,20 @@ def authenticate_admin
...
+  private
+
+  def no_access_guest
+    redirect_to new_user_session_path
+    no_access_flash
+  end
+
+  def no_access_user
+    redirect_to root_path
+    no_access_flash
```
nicholasjhenry: While this does work, it's more idiomatic to set the flash before doing the redirect.

app/controllers/admin/application_controller.rb
```diff
@@ -16,4 +17,20 @@ def authenticate_admin
   # def records_per_page
   #   params[:per_page] || 20
   # end
+
+  private
+
+  def no_access_guest
+    redirect_to new_user_session_path
```
nicholasjhenry: Perhaps method constructed with a verb would be more appropriate e.g.

```
no_access_guest -> handle_unauthenticated_user
no_access_user -> handle_unauthorized_user
```

app/controllers/admin/application_controller.rb
```diff
@@ -16,4 +17,20 @@ def authenticate_admin
...
+    no_access_flash
+  end
+
+  def no_access_user
+    redirect_to root_path
+    no_access_flash
+  end
+
+  def no_access_flash
+    flash[:alert] = "Sorry, you just don't have access rights!"
```
nicholasjhenry: Should we have a different flash message depending on if they were unauthenticated vs. unauthorized?

spec/features/authentication_spec.rb
```diff
@@ -1,28 +1,55 @@
...
 
-    click_link 'Sign out'
-    expect(page).not_to have_content('Sign out')
-    expect(page).to have_content('Signed out successfully')
+      click_link "Sign out"
+      expect(page).to have_content("Signed out successfully")
+    end
+
+    scenario "Can not access admin area" do
+      sign_in(user)
```
nicholasjhenry: I like how you have separated the lines exercising the app from the expectations in the scenarios above. Can we do the same for the scenario below. Definitely makes it easy to read.

spec/features/authentication_spec.rb
```diff
@@ -1,28 +1,55 @@
...
+      expect(page).to have_content("Sorry, you just don't have access rights!")
+    end
   end
 
-  scenario "Signs me out" do
-    sign_in(user)
+  context "Regular user" do
+    scenario "Can sign in and out" do
+      sign_in(user)
+      expect(page).to have_content("Signed in successfully")
```
nicholasjhenry: Expectation on text can be brittle (i.e. change the test in the UI and you've broken a spec). Matching a word in an element can help with this. From:

```ruby
expect(page).to have_content("Signed in successfully")
```

to:

```ruby
expect(page).to have_css('.alert-success', text: "success")¬
```

Honestly, though -- I'm happy to just check on an element to see if it exists; that might make a few folks nervous:

```ruby
expect(page).to have_css('.alert-success')¬
```

but you can create a nice RSpec matcher for it:

```ruby
expect(page).to have_success_notification¬
```

I'll let you decide how far you want to take that advice :8ball:

nicholasjhenry:
Thank you for contributing this @karimmtarek. I've given this a thorough review as I know you would want nothing less.

For those who don't know -- @karimmtarek and I spent many a Friday afternoon talking about code, looking at code and even writing a bit of code ;-)

karimmtarek:
@nicholasjhenry thanks a lot for the thorough review, I really appreciate it.
I really miss those days @nicholasjhenry, I was super lucky to have you as a mentor :heart_eyes:

benichu:
@karimmtarek keep on contributing and those days won't be over :smile:

karimmtarek:
@benichu thats what I'm planing to do ;)

benichu:
Looks good, do you want to squash your commits into one? 
See comment here: https://github.com/montrealrb/Montreal.rb/pull/137#issuecomment-162260455

I'll be happy to merge to master after this.

karimmtarek:
@benichu done :+1:

Issue #83, opened 2015-11-27
Allow the body of news items and events to be written in markdown


itsWill:
It would be cool if for news items and events we could write the body in markdown to give our posts some :star: :gem: :star: :gem: :star: :gem: :star:.  Thinking we could implement this using something like [redcarpet](https://github.com/vmg/redcarpet)

cawel:
Good idea! I believe @nicholasjhenry told me that was indeed in the pipeline. Pending the core team's approval (e.g. by tagging the issue as a 'feature'), I'd be happy to start on that one.

itsWill:
I've implemented something like this before, so let me know if you need some pointers on how to tackle this one.

benichu:
:+1:

@cawel do you want to take care of it? Also, remember that the markdown storage/display behaviour should be shared by a lot of our models (anywhere we have a `description` or `body` fields I believe, source: https://github.com/montrealrb/Montreal.rb/issues/47#issuecomment-158508238)

cawel:
@benichu Sure thing :+1:  I will start on that tomorrow Tuesday.

benichu:
Cool stuff. I invited you to the contributor's team, so you don't have to fork the repository but can directly create branches off the repository. Tell me if you need any help.

nicholasjhenry:
One additional feature on Events and NewsItems would be to include a `filter` attribute to specify if it's in `markdown` or `html`. This would help with rendering legacy content imported from Wordpress.

cawel:
@itsWill Thanks for the offer. I've been writing my own notes – on various topics – using the awesome markdown syntax for many years now. I should be good! But I'll be looking forward to hear what you have to say in a possible code review ;)

@benichu Thanks for making me a contributor :bow:

@nicholasjhenry Happy if I make your `filter` attribute an enum with 2 possibles options (`markdown` and `html`)? I would speculate that this attribute will be useful initially when rendering the legacy content in its original `html`. But once all that legacy content will be converted to markdown (some months from now), the idea is to get rid of that attribute and only support markdown for _all_ user-defined content.

Taking a look at the [object model diagram](https://github.com/montrealrb/Montreal.rb/issues/47#issuecomment-158508238), my understanding is that "markdown support" is required for the following models:

* `Event`
* `Job`
* `NewsItem`
* `Organization`
* `Page`
* `Talk`

So just to double check, `Event` and `NewsItem` are the only models which should have that special `filter` attribute?

[...]

PR #98, opened 2015-12-03
[chore] Fixes #42 - Import old wordpress data and convert to markdown…

### LINKED ISSUE #42: Migrate data from old site
I'm assuming we are doing this?

- [x] source dump of MySQL DB
- [x] import posts to news
  - [x] convert html to markdown

benichu:
Left TODOS:
- [x] document execution of rake command on heroku
- [x] cleanup NewsItem model
- [x] controller tests for news_items#show and news_items#index
- [x] unit tests `.published`

PROBLEMS:
- Overriding thoughtbot/administrate does not work as advertised

[Review on Reviewable](https://reviewable.io/reviews/montrealrb/montreal.rb/98)

### DIFF
--- a/app/controllers/news_items_controller.rb
+++ b/app/controllers/news_items_controller.rb
@@ -1,9 +1,23 @@
 class NewsItemsController < ApplicationController
   def index
-    @news_items = NewsItem.published.to_a
+    @news_items = news_item_scope.to_a
   end
 
+  # TODO: See Issue #151
+  # If `params[:slug]` is present, it's *most probably* coming from a
+  # wordpress legacy link.
+  # When the legacy data was imported, we also stored that original `slug`
+  # and this is why we first try to load an event_item using that param.
+  #
   def show
-    @news_item = NewsItem.published.find(params[:id])
+    @news_item = news_item_scope.where(slug: params[:slug]).first ||
+                 news_item_scope.where(id: params[:id]).first ||
+                 raise(ActiveRecord::RecordNotFound)
+  end
+
+  private
+
+  def news_item_scope
+    NewsItem.published
   end
 end
--- a/app/dashboards/news_item_dashboard.rb
+++ b/app/dashboards/news_item_dashboard.rb
@@ -12,7 +12,7 @@ class NewsItemDashboard < Administrate::BaseDashboard
     id: Field::Number,
     title: Field::String,
     body: Field::Text,
-    state: Field::String,
+    state: EnumField,
     published_at: Field::DateTime,
     created_at: Field::DateTime,
     updated_at: Field::DateTime,
@@ -26,11 +26,9 @@ class NewsItemDashboard < Administrate::BaseDashboard
   COLLECTION_ATTRIBUTES = [
     :id,
     :title,
-    :body,
     :state,
-    :published_at,
-    :created_at,
     :updated_at,
+    :published_at,
   ]
 
   # SHOW_PAGE_ATTRIBUTES
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -75,7 +75,10 @@ ActiveRecord::Schema.define(version: 20151204174632) do
     t.string   "title"
     t.datetime "created_at",   null: false
     t.datetime "updated_at",   null: false
+    t.string   "slug"
   end
+
+  add_index "news_items", ["slug"], name: "index_news_items_on_slug", using: :btree
 
   create_table "organization_translations", force: :cascade do |t|
     t.integer  "organization_id", null: false
--- a/lib/tasks/database.rake
+++ b/lib/tasks/database.rake
@@ -1,0 +1,74 @@
+require "mysql2"
+require "yaml"
+require "nokogiri"
+require "reverse_markdown"
+
+namespace :database do
+  namespace :legacy do
+    # NOTE to run remotely on heroku:
+    # Source: https://devcenter.heroku.com/articles/rake
+    #
+    # $ heroku run rake database:legacy:import_news
+    #
+    desc "Import news records to database"
+    task :import_news => :environment do
+      STDOUT.puts "This will destroy your news_items table. Enter 'Y' to confirm: [y/N]"
+      input = STDIN.gets.chomp
+      raise "Aborting import. You entered #{input}" unless input.downcase == "y"
+
+      NewsItem.destroy_all
+      records = YAML.load_file "#{Rails.root}/db/legacy.yml"
+      records.each do |data|
+        news_item = NewsItem.new
+        published_at = data["post_date"]
+
+        news_item.state        = published_at > DateTime.new(2015, 1, 1).beginning_of_year ? :published : :archived
+        news_item.title        = data["post_title"]
+        news_item.body         = data["post_content"]
+        news_item.published_at = data["post_date"]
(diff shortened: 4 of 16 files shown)


benichu:
waiting for #83 to be merged before finishing it up

benichu:
@montrealrb/organizers If anyone of you is interested to do a code-review and merge that PR, it takes care of the following:

- Generate a wordpress database dump in db/legacy.yml
- Create a rake task to import that legacy data in the news_items table
- Add some unit and controller specs for NewsItem
- Deal with legacy urls: /YYYY/mm/a-wordpress-slug

Note: you will noticed that I imported all the posts from wordpress (hopefully) but that I only marked the 10 most recent, because some manual markdown cleanup work will need to be done, and I think we should focus on the most recent posts for launch and after that work our way back...

benichu:
code reviewer, please assign the PR to yourself...

benichu:
Will also Fix #99

app/dashboards/news_item_dashboard.rb
```diff
@@ -26,11 +26,9 @@ class NewsItemDashboard < Administrate::BaseDashboard
   COLLECTION_ATTRIBUTES = [
     :id,
     :title,
-    :body,
     :state,
-    :published_at,
-    :created_at,
     :updated_at,
+    :published_at,
```
mdeloupy: @benichu This array ends with a comma, is that normal ? Coming from you I guess it is, but I've never seen this behavior.

mdeloupy: Got my answer : https://github.com/bbatsov/ruby-style-guide/issues/76

lib/tasks/database.rake
```diff
@@ -0,0 +1,74 @@
...
+
+namespace :database do
+  namespace :legacy do
+    # NOTE to run remotely on heroku:
+    # Source: https://devcenter.heroku.com/articles/rake
+    #
+    # $ heroku run rake database:legacy:import_news
+    #
+    desc "Import news records to database"
+    task :import_news => :environment do
```
mdeloupy: Hash rocket

lib/tasks/database.rake
```diff
@@ -0,0 +1,74 @@
...
+        rescue => e
+          puts news_item.inspect
+        end
+      end
+    end
+
+    # NOTE: This has normally been run by someone with access to the wordpress mysql database dump
+    # The result is `db/legacy.yml`
+    desc "Generate a yml files with legacy news records from wordpress mysql DB"
+    task :dump => :environment do
```
mdeloupy: Hash rocket

mdeloupy:
@montrealrb/organizers I did a quick review, but don't want to take the responsibility of merging this... Anyone else up for it ?

app/controllers/news_items_controller.rb
```diff
@@ -1,9 +1,17 @@
...
   def index
-    @news_items = NewsItem.published.to_a
+    @news_items = news_item_scope.to_a
   end
 
   def show
-    @news_item = NewsItem.published.find(params[:id])
+    @news_item = news_item_scope.where(slug: params[:slug]).first ||
+                 news_item_scope.where(id: params[:id]).first ||
+                 raise(ActiveRecord::RecordNotFound)
```
nicholasjhenry: @benichu Could we extract as a method and pushed into the model? e.g.

```ruby
NewsItem.fetch_published!(params[:id], params[:slug])
```

This allow to document the method, as I'm unsure why this is required.

db/schema.rb
```diff
@@ -16,6 +16,25 @@
   # These are extensions that must be enabled in order to support this database
   enable_extension "plpgsql"
 
+  create_table "companies", force: :cascade do |t|
```
nicholasjhenry: @benichu can you please do a `rake db:migrate:reset` to update your schema? These tables should not be in the schema.

app/controllers/news_items_controller.rb
benichu: Sure. The reason why it's required is to still resolved old wordpress links (ex: if someone clicks on an post url from google)
I'll extract that logic into a tested and documented class method tonight or later today...

db/schema.rb
benichu: will do.

lib/tasks/database.rake
```diff
@@ -0,0 +1,74 @@
...
+    task :import_news => :environment do
+      STDOUT.puts "This will destroy your news_items table. Enter 'Y' to confirm: [y/N]"
+      input = STDIN.gets.chomp
+      raise "Aborting import. You entered #{input}" unless input.downcase == "y"
+
+      NewsItem.destroy_all
+      records = YAML.load_file "#{Rails.root}/db/legacy.yml"
+      records.each do |data|
+        news_item = NewsItem.new
+        published_at = data["post_date"]
```
nicholasjhenry: @benichu Since there's a couple of changes that need to be made, how would you feel call #fetch on these?

```
 data.fetch("post_date")
```

I like exceptions thrown when the key I expect doesn't exist.

lib/tasks/database.rake
```diff
@@ -0,0 +1,74 @@
...
+
+        news_item.state        = published_at > DateTime.new(2015, 1, 1).beginning_of_year ? :published : :archived
```
nicholasjhenry: I added the state `archived` to the domain model. Perhaps we don't actually need this. Thoughts?

lib/tasks/database.rake
benichu: `published_at` could be nil, if the original post is a draft (which will not happen with the collection I am parsing, as the original query generated only published posts with a date). So it's not really necessary to catch the non-existence of that key.

NOTE: It's a one-off task anyway

lib/tasks/database.rake
benichu: The only reason why I am doing this, is because I wanted to automatically mark as published the 2015 posts, leaving us time to clean up old posts one by one.

app/controllers/news_items_controller.rb
nicholasjhenry: Gotcha, we probably should do a redirect to the correct URL, but let's do that as a separate PR.

lib/tasks/database.rake
nicholasjhenry: Not quite what I meant -- the key should exists, but the value might be nil :-) I should have clarified that this would be for all keys we're accessing in the hash. But happy to move on given it's a one-off task.

lib/tasks/database.rake
nicholasjhenry: Perfect.

nicholasjhenry:
Awesome, @benichu! I ran this locally -- looks like we got us a site :-)

benichu:
Cool, I'll do a couple of the fixes you mentioned and we are good to merge, talk to you later...

app/controllers/news_items_controller.rb
benichu: You are right, just created a ticket for that #151
Until then, I am also going to leave the object loading methods as-is in the controller, I'll just add a comment to the method.

benichu:
@nicholasjhenry rebased, I'll merge tomorrow morning.

sophiedeziel:
Haha you have to rebase again :stuck_out_tongue:

Issue #94, opened 2015-12-02
Error when try to create an event with an empty form


karimmtarek:
Below is the error I got when I tried to create an event with an empty form
<img width="1223" alt="screen shot 2015-12-01 at 9 52 40 pm" src="https://cloud.githubusercontent.com/assets/5586031/11520958/f0b91f28-9875-11e5-8bf7-01d43d8d4add.png">

sophiedeziel:
Oh, this is database-level validation. We should shield it by having active record validations

itsWill:
this is to do with it violating the [schema](https://github.com/montrealrb/Montreal.rb/blob/master/db/schema.rb#L34). We can prevent this with validations at the model.

cawel:
Since that issue was really simple, I pushed a fix to the `master` branch directly. See [commit](https://github.com/montrealrb/Montreal.rb/commit/104bda6a34e5b6dd8957918e6c1b66bf7691a73b).

cawel:
As well as some specs for that `Event` model (see [commit](https://github.com/montrealrb/Montreal.rb/commit/b8701b6c5e0b3362390771626ee78e88da2d72f8)).

nicholasjhenry:
thank you for handling this @cawel

PR #100, opened 2015-12-03
Creating an event with default values explodes in the admin interface

cawel:
This PR is a pretext to close issue #94 (which already has its fix in the `master` branch). In addition to that, I took the opportunity to make `Event#title` a required field.

Note that I modified the original "events" migration directly, as I don't think we have proper data in the database just yet. If this is causing any kind of problem, I'm happy to revert those changes and create a brand new migration expressly for that.

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -15,6 +15,7 @@ class Event < ActiveRecord::Base
   # suggestion from Nicholas
   to_param :title
 
+  validates :title, presence: true
   validates :starts_at, presence: true
 
   def self.published
--- a/db/migrate/20140929233719_create_events.rb
+++ b/db/migrate/20140929233719_create_events.rb
@@ -5,7 +5,7 @@ class CreateEvents < ActiveRecord::Migration
       t.datetime :starts_at, null: false
       t.timestamps
     end
-    Event.create_translation_table! :title => :string, :introduction => :text, :conclusion => :text
+    Event.create_translation_table! title: :string, introduction: :text, conclusion: :text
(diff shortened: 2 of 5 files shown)


nicholasjhenry:
> Note that I modified the original "events" migration directly, as I don't think we have proper data in the database just yet. If this is causing any kind of problem, I'm happy to revert those changes and create a brand new migration expressly for that.

Normally I would be totally on board with this -- the only consideration is ensuring contributors do a `rake db:migrate:reset` when submitting a PR as there schema.rb maybe generated without that null constraint and ends up reverting the change in the schema.rb. I've seen a bit of "bouncing" already with timestamps that now have null constraints on these.

Not so important, but we would need to reset the production DB on Heroku after the merge as it's deployed automatically.

cawel:
@nicholasjhenry I also saw this annoying timestamps bouncing. Hence why we need the most trusted people to double-check things before merging them into `master` ;)

And yeah, resetting the prod db should be easy. We just have to remember to do it.

Just to be sure: you started your comment using the word "normally" (which could mean that the current circumstances would qualify as an exception to the rule), but I don't think you want me to change anything on that issue as it stands now?

nicholasjhenry:
Yes, I was a little vague on the call to action on that one. I started off against the idea (of modifying the migration directly) and then talked myself out of it. But having said that...

If you don't mind -- create a new migration for it. Any thing we can do to reduce the delay of merging PR's is a win. Thanks!

cawel:
Done !

Issue #107, opened 2015-12-04
Creating an `Event` fails with no location


nicholasjhenry:
<img width="1111" alt="screenshot 2015-12-04 11 20 58" src="https://cloud.githubusercontent.com/assets/40253/11594702/85494320-9a79-11e5-83e8-ec22d7b13d5a.png">

cawel:
This is waiting on #116 to be resolved.

sophiedeziel:
#116 is merged

PR #116, opened 2015-12-04
Events: make 'location_id' a required field

cawel:
Fix for issue #107.

[Review on Reviewable](https://reviewable.io/reviews/montrealrb/montreal.rb/116)

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -17,13 +17,10 @@ class Event < ActiveRecord::Base
 
   validates :title, presence: true
   validates :starts_at, presence: true
+  validates :location, presence: true
 
   def self.published
     order(starts_at: :desc)
-  end
-
-  def location
-    super || TBALocation.new
   end
 
   def title_with_date
--- a/app/models/location.rb
+++ b/app/models/location.rb
@@ -3,26 +3,3 @@ class Location < ActiveRecord::Base
 
   validates :name, presence: true
 end
-
-class TBALocation
-  def name
-    'TBA'
-  end
-
-  def address
-    'TBA'
-  end
-
-  def directions
-    ''
-  end
-
-  def description
-    ''
-  end
-
-  def url
-    ''
-  end
-
-end
--- a/db/migrate/20151204174632_make_events_location_id_required.rb
+++ b/db/migrate/20151204174632_make_events_location_id_required.rb
@@ -1,0 +1,9 @@
+class MakeEventsLocationIdRequired < ActiveRecord::Migration
+  def up
+    change_column :events, :location_id, :integer, null: false
+  end
+
+  def down
+    change_column :events, :location_id, :integer, null: true
+  end
+end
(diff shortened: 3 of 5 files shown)


app/models/event.rb
```diff
@@ -16,6 +16,7 @@ class Event < ActiveRecord::Base
   to_param :title
 
   validates :starts_at, presence: true
+  validates :location_id, presence: true
```
nicholasjhenry: @cawel Can we get the validation on `:location` rather than the id, please?

Ernie Miller makes a good case for this in his talk "An Intervention for ActiveRecord by Ernie Miller" -- see the segment "Deep Thoughts on Validations" 

https://youtu.be/yuh9COzp5vo?t=1067

cawel: @nicholasjhenry Sure thing. I ended up watching the whole talk. Funny one :)

As for the requested change, almost a year ago @mcgain has created a `location` method in the `Event` class, in effect overriding its `location` relationship like this:

```ruby
def location
  super || TBALocation.new
end
```

And it's giving me problems since that `TBALocation` object (which stands for "to-be-announced location) is not an `ActiveRecord` object and it explodes when attempting to validate the `location` relationship (`undefined method 'marked_for_destruction?' for #<TBALocation:0x007fa6eb88b050>`). 

Unless someone gives me a good reason to keep that `TBALocation` class, I'd like to remove it entirely.

On top of that, I would like to [renegotiate the requirements](https://vimeo.com/4552747): do we really need to make the location _required_ for an event? What if you want to create (i.e. persist it in db) an event today, but you still don't know where it's going to take place (the concept of a "to-be-announced" location). For that reason, I think there should *not* be a presence validation on the `location` (and naturally neither on `location_id`). What do you think?

sophiedeziel:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion.

---

*[app/models/event.rb, line 19 \[r1\]](https://reviewable.io:443/reviews/montrealrb/montreal.rb/116#-K4zFd82wVEU21p0XlXG-r1-19:-K58s6czjR0xNI5VtEt1:-2117202296) ([raw file](https://github.com/montrealrb/montreal.rb/blob/bafb0eb6e83253a51f72f6d00b366790304bd342/app/models/event.rb#L19)):*
For a TBA location, since only the name is required for a location, we can create one in the database named "TBA". We don't need a class and fancy logic to tell people that it is to be announced! Once the location is ready to be announced, we update the record with the correct location ;)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/montrealrb/montreal.rb/116)*

sophiedeziel:
Reviewed 1 of 4 files at r1, 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/montrealrb/montreal.rb/116)*

sophiedeziel:
fixes #107.

sophiedeziel:
fixes #107

cawel:
:+1: Bye bye `TBALocation` !

Issue #117, opened 2015-12-04
Add validation to news items


nicholasjhenry:
body, title, state should be required.

PR #125, opened 2015-12-04
Fixed Issue #117 Add Validation to News Items

AndyObtiva:
Fixed https://github.com/montrealrb/Montreal.rb/issues/117
As part of that, I added news item rspec tests and altered display of state field in the admin dashboard for news item to show a dropdown for the state enum. I added shoulda-matchers as part of my work on the tests. I hope you don't mind. I don't like to use them for every validation test, but only for the most simple and mundane boilerplate ones. The rest, I wrote standard rspec tests for.

This is my first contribution, so I'm new and can benefit from criticism and pointing in the right direction given I have not worked with this team before. 

Thanks for the opportunity and cheers.

[Review on Reviewable](https://reviewable.io/reviews/montrealrb/montreal.rb/125)

### DIFF
--- a/app/models/news_item.rb
+++ b/app/models/news_item.rb
@@ -17,14 +17,22 @@ class NewsItem < ActiveRecord::Base
   scope :published, -> { where(state: :published).order(published_at: :desc) }
 
   # Validations
-  validates :published_at, presence: true, if: -> { state.try(:published?) }
-  validates :title, presence: true, length: {maximum: 255}
-  validates :state, presence: true, inclusion: { in: AVAILABLE_STATES }
-  validates :body, presence: true, length: {maximum: 4000}
+  validates :published_at,
+            presence: true,
+            if: -> { state.try(:published?) }
+  validates :title,
+            presence: true,
+            length: { maximum: MAX_STRING_COLUMN_LENGTH }
+  validates :body,
+            presence: true,
+            length: { maximum: MAX_TEXT_COLUMN_LENGTH }
+  validates :state,
+            presence: true,
+            inclusion: { in: STATES }
 
   # Class methods
-  AVAILABLE_STATES = %w(draft archived published)
-  enumerize :state, in: AVAILABLE_STATES
+  STATES = %w(draft published archived)
+  enumerize :state, in: STATES
   to_param :title
 
   # Instance methods
--- a/app/models/news_item.rb
+++ b/app/models/news_item.rb
@@ -17,11 +17,14 @@ class NewsItem < ActiveRecord::Base
   scope :published, -> { where(state: :published).order(published_at: :desc) }
 
   # Validations
-  validates :published_at, presence: true, if: -> { state.published? }
-  validates_presence_of :title, :state
+  validates :published_at, presence: true, if: -> { state.try(:published?) }
+  validates :title, presence: true, length: {maximum: 255}
+  validates :state, presence: true, inclusion: {in: AVAILABLE_STATES, message: "must be one of the following: #{AVAILABLE_STATES.join(', ')}"}
+  validates :body, presence: true, length: {maximum: 4000}
 
   # Class methods
-  enumerize :state, in: [:draft, :archived, :published]
+  AVAILABLE_STATES = %w(draft archived published)
+  enumerize :state, in: AVAILABLE_STATES
   to_param :title
 
   # Instance methods
--- a/app/models/news_item.rb
+++ b/app/models/news_item.rb
@@ -13,6 +13,9 @@ class NewsItem < ActiveRecord::Base
   # Extends
   extend Enumerize
 
+  # Constants
+  STATES = %w(draft published archived)
+
   # Scopes
   scope :published, -> { where(state: :published).order(published_at: :desc) }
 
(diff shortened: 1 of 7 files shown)


mathieugagne:
@AndyObtiva Thanks for the pull request! Would you mind having a look at Travis? [https://travis-ci.org/montrealrb/Montreal.rb/builds/94991453](https://travis-ci.org/montrealrb/Montreal.rb/builds/94991453)

It's complaining about missing `NewsItem#body` on controller specs.

AndyObtiva:
Thanks for the note Mathieu. I fixed it and improved the code style to make Hound happy.

app/models/news_item.rb
```diff
@@ -1,7 +1,10 @@
...
-  enumerize :state, in: [:draft, :archived, :published]
+  AVAILABLE_STATES = %w(draft archived published)
+  enumerize :state, in: AVAILABLE_STATES
   scope :published, -> { where(state: :published).order(:published_at) }
-  validates :published_at, presence: true, if: -> { state.published? } 
-  validates_presence_of :title, :state
+  validates :published_at, presence: true, if: -> { state.try(:published?) }
+  validates :title, presence: true, length: { maximum: 255 }
+  validates :state, presence: true, inclusion: { in: AVAILABLE_STATES }
+  validates :body, presence: true, length: { maximum: 4000 }
```
mathieugagne: Out of curiosity, why 4000?

spec/models/news_item_spec.rb
```diff
@@ -1,5 +1,61 @@
...
+require "rails_helper"
 
 RSpec.describe NewsItem, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+  describe "validation" do
+    it { should validate_presence_of(:title) }
+    it { should validate_length_of(:title).is_at_most(255) }
+    it { should validate_presence_of(:body) }
+    it { should validate_length_of(:body).is_at_most(4000) }
+    context "for state" do
```
mathieugagne: `shoulda-matchers` has a nifty helper for validating inclusions. Since you already have it above, is there another reason to do it explicitly?

AndyObtiva: Yes, I wrote it with the shoulda-matcher originally, but it kept failing because it would try to test that the enum state field can accept a symbol input for its setter, but would return a string output. It didn't like the asymmetry and complained about it as bad API. I wish I'd discover a way to relax its strictness, but then I realized it's not worth spending too much time forcing it to work for a few validations, and just wrote my version of inclusion testing.

app/models/news_item.rb
AndyObtiva: In my experience, some databases support up to 4000 characters for "text" columns (like "news_items"."body"), so I just did that reflexively thinking it was good enough a limit for now (I didn't bother to check what the limits are in this project's postgres, but hey 4000 is pretty high, no?! :) )

spec/models/news_item_spec.rb
cawel: Well, that might just be another argument for [dumping the `Enumerize` gem](https://github.com/montrealrb/Montreal.rb/issues/128) then ;)

app/models/news_item.rb
AndyObtiva: I just double-checked and postgres has no limit on its text field length: http://www.postgresql.org/docs/9.1/static/datatype-character.html

I think I reflexively did that also because it's a security recommendation to limit intake of information from users (e.g. preventing users from providing endless data that would consume all server resources).

In any case, I just checked the live site posts, and although only one out of all hit the 5000 char limit, if I take their HTML content into account, they get much higher than that. 

I'll bump up that limit to 64k (65536 bytes) just to be safe.

sophiedeziel:
Reviewed 3 of 6 files at r1, 1 of 3 files at r2, 3 of 3 files at r3.
Review status: 7 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[Gemfile, line 60 \[r3\]](https://reviewable.io:443/reviews/montrealrb/montreal.rb/125#-K59A_8N1dZheMPQXWt2:-K59A_8N1dZheMPQXWt3:355366674) ([raw file](https://github.com/montrealrb/montreal.rb/blob/f54fac5092aed79530ed1d5de5288944d376d9f9/Gemfile#L60)):*
I love shoulda-matchers

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/montrealrb/montreal.rb/125)*

sophiedeziel:
Reviewed 1 of 6 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/montrealrb/montreal.rb/125)*

spec/models/news_item_spec.rb
```diff
@@ -1,5 +1,67 @@
-require 'rails_helper'
+require "rails_helper"
 
 RSpec.describe NewsItem, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
+  describe "validation" do
+    it { should validate_presence_of(:title) }
```
nicholasjhenry: A heads up that you can use the `expect` syntax for shoulda methods. Keeps the expectations uniformed.

```ruby
it { is_expected.to validate_presence_of(:title) }
```

AndyObtiva: Thanks for the note. Will submit a revision.

AndyObtiva:
Rebased.

sophiedeziel:
fixed by #187

PR #134, opened 2015-12-05
Added .ruby-gemset

AndyObtiva:
Configured rvm gemset to scope project gems independently of other Ruby projects. For example, that makes it possible to configure same gem versions for different projects, but with different native binding configurations or CPU architectures (e.g. x86 vs x64). Also, it makes it easier to delete all project gems from machine when removing project or upgrading to a new version of Ruby, without disturbing other project gems that might be dependent on older version of Ruby (using "rvm gemset empty" command).

[Review on Reviewable](https://reviewable.io/reviews/montrealrb/montreal.rb/134)

### DIFF
--- a/.ruby-gemset
+++ b/.ruby-gemset
@@ -1,0 +1,1 @@
+Montreal-rb


mathieugagne:
@montrealrb/organizers I added a discussion label to this one and #135. I don't use either of them (rvm or pow) but don't have a problem with dotfiles.

benichu:
I wonder if we should just update our `bin/setup` script (and document) how bundler can vendorize gem dependencies per project, like so (if I remember correctly):

```bash
bundle install --path vendor
echo 'vendor/ruby' >> .gitignore
``` 
This would allow us to be consistent whether our contributors user rvm, rbenv or other methods to manage their ruby version...

sophiedeziel:
Dotfiles are mostly harmless but the project should be "ruby manager agnostic"... In the same way I don't commit dotfiles for the IDE project by adding them to my global gitignore..

nicholasjhenry:
Thank you for raising this discussion @AndyObtiva. As @sophiedeziel pointed out. We're going to keep dotfiles to a minimum (only added for integration services etc) and consider this a personal config option.

PR #135, opened 2015-12-05
Added .powrc 

AndyObtiva:
Added pow configuration to leverage RVM in supporting POW multi-project friendly-URL local hosting (e.g. local site works as http://montreal.rb.dev instead-of or in-addition-to http://localhost:3000; check http://pow.cx for details)


nicholasjhenry:
Thanks for raising this @AndyObtiva. I won't merge this into master based on our reasons outline here:

https://github.com/montrealrb/Montreal.rb/pull/134#issuecomment-163802391

Issue #93, opened 2015-12-02
Jobs needs to be validated


karimmtarek:
- A Job can be created with an empty form, validation should be added to prevent this.

PR #136, opened 2015-12-05
Added jobs validation

AndyObtiva:
Implemented https://github.com/montrealrb/Montreal.rb/issues/93 adding job model/form validations

[Review on Reviewable](https://reviewable.io/reviews/montrealrb/montreal.rb/136)

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -1,6 +1,15 @@
 class Job < ActiveRecord::Base
   extend Enumerize
-  STATUSES = [:draft, :published, :archived]
-  enumerize :state, in: STATUSES
+  STATES = %w(draft published archived)
+  enumerize :state, in: STATES
   belongs_to :organization
+  validates :title,
+            presence: true,
+            length: { maximum: MAX_STRING_COLUMN_LENGTH }
+  validates :description,
+            presence: true,
+            length: { maximum: MAX_TEXT_COLUMN_LENGTH }
+  validates :state,
+            presence: true,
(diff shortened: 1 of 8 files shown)


app/models/job.rb
```diff
@@ -1,6 +1,11 @@
 class Job < ActiveRecord::Base
   extend Enumerize
-  STATUSES = [:draft, :published, :archived]
+  STATUSES = %w(draft published archived)
   enumerize :state, in: STATUSES
```
cawel: Smell: we're using 2 words ('status' and 'state') to refer to the same concept.

Maybe we could simply say this?

```ruby
STATES = %w(draft published archived)
enumerize :state, in: STATES
```

AndyObtiva: Good point. I just revised it to say STATES.

nicholasjhenry:
Just a heads up @AndyObtiva that this requires a rebase. Cheers!

AndyObtiva:
OK, rebased from trunk.

steakunderscore:
@AndyObtiva Mind if I update this patch? Or would you be able to? I was also wanting to use shoulda-matchers.

AndyObtiva:
Go ahead if you like. I can't stop you if GitHub gives you and everybody access.

steakunderscore:
@AndyObtiva Thanks, but I still don't want to be stepping on ya toes!

sophiedeziel:
fixed by #177

Issue #108, opened 2015-12-04
Remove `Translation IDS` from admin `Location` screen


nicholasjhenry:
<img width="943" alt="screenshot 2015-12-04 11 26 43" src="https://cloud.githubusercontent.com/assets/40253/11594973/fcd1fa6c-9a7a-11e5-86d5-a7ab76d63a31.png">

cawel:
Done ! PR #107 created.

PR #137, opened 2015-12-05
Locations: remove 'translation ids' field in admin interface

cawel:
Fix issue #108.

### DIFF
--- a/app/dashboards/location_dashboard.rb
+++ b/app/dashboards/location_dashboard.rb
@@ -37,7 +37,6 @@ class LocationDashboard < Administrate::BaseDashboard
   # an array of attributes that will be displayed
   # on the model's form (`new` and `edit`) pages.
   FORM_ATTRIBUTES = [
-    :translations,
     :name,
     :address,
     :url,


benichu:
I don't think the issue related to that PR have been closed. I will close it.
Way to do it: https://help.github.com/articles/closing-issues-via-commit-messages/

cawel:
@benichu Sweet. Will try that next time!

benichu:
@cawel there is also a discussion that we are having about optimizing the contribution process, here: https://github.com/montrealrb/playbook/issues/1

It will probably involve some kind of commit squashing step before merging, where the commit message could be formatted in a way that our master git history is clean and focused but also would make that "auto-closing" of issues work.

It could be something like:

```
[feature] Fixes #111 - Description of the feature 
Body of commit with more context and information...
```

Here is an example of one my recent commits: https://github.com/montrealrb/Montreal.rb/commit/2aca4d14e64af423314e38f17f32f38fe44ff37e

cawel:
@benichu I see. I can't say I'm convinced about squashing the commits. I personally prefer when changes are incremental, even if they are related to the same issue. For instance, if the issue at hand is quite big in scope, then you would have _one single commit_ impacting a lot of different things in several places in the codebase. That does not feel right to me. 

And if for some reason you'd want to merge in only half of those changes after all, that would not be an easy task. To push the argument to the extreme, that means the `master` branch should have as many commits as we have resolved issues/features. Another thing that does not feel right to me. I think it's in git's DNA to have relevantly-bundled changes through time, regardless of their context. Another way to look at it: git is a tool that should serve the developer more than the project manager.

On the other hand, squashing commits and incorporating that "Fixes issue X" bit in the git message feels like wrapping things up in a neat way. I think it's an ideal practice for a bug fix; less so for a feature or a significant change.

benichu:
I guess I have more the project manager's point of view :)
The way I look at it and apply every day at work, dealing with a lot of projects:

1. An issue is created, and that issue is a problem that we, as developers, need to solve (It can be a bugfix or a feature)
1. If that issue is a big feature, it will most likely have to be split into other issues beforehand, reducing the scope and potentially the complexity of solving that problem (but that would be the subject of a different discussion altogether) 
1. As soon as a developer takes charge of a given issue, a branch and pull request is created (normally named accordingly, for example: `feature-bt-111-description-of-feature`, I usually prepend the title of my pull request with `[wip]` to explicitely mark that the pull request is not ready to be merged at any cost (using gitlab is pretty useful when you apply that convention, as it prevents merging)
1. That pull request becomes a sandbox for the developer, as well as way to remotely backup the work being done... I personally encourage my coworkers to commit and push very often to always have constant discussion about work in progress (always keeping an eye on the commit feed...)
1. Once the developer is ready for a code-review, he can notify reviewers via a comment, a reviewer then assign the PR to himself, comments, asks for changes, etc...
1. Once the pull request is good, the developer performs the rebase and squash into 1 historic commit (or many if the PR was solving several issues, but that should be pretty rare...)
1. The code-reviewer can now merge with a "no fast-forward", and you benefit from having the final history in master showing 2 commits: who did the work, and who accepted it.
1. The life of your project becomes very easily traceable through its commits, changelogs are easier to write, etc...

This process is a good balance between granularity of commits while solving a problem, facilitating discussion (usually useful at a single point in time) and the need for a higher level log of actions attached to external references such as bugs or feature stories (useful for the lifetime of the project).

I would __never__ try to force anyone to use that process in the montrealrb context, but this works __very__ well in a professional software development context, with many concurrent projects, collaborators and non-technical stakeholders...

We can definitely discuss this further during the next montrealrb meetup, this is one of my favorite subjects :beers:

Issue #106, opened 2015-12-04
Simplify Event attributes


nicholasjhenry:
Can `introduction` and `conclusion` be simplified to just `description`? Any need for these two attributes?

benichu:
:+1:

nicholasjhenry:
Great -- this is open for anyone to pick up.

itsWill:
I can pick this one up as well as I work thorough https://github.com/montrealrb/Montreal.rb/issues/105

PR #138, opened 2015-12-06
Added body to events removed introduction/conclusion and properly display events on the show page

itsWill:
This fixes https://github.com/montrealrb/Montreal.rb/issues/106 and the outstanding tasks in https://github.com/montrealrb/Montreal.rb/issues/14

@nicholasjhenry

[Review on Reviewable](https://reviewable.io/reviews/montrealrb/montreal.rb/138)

### DIFF
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -15,6 +15,7 @@ class Event < ActiveRecord::Base
   validates :state,
             presence: true,
             inclusion: { in: STATES }
+  validates :body, presence: true
 
   enumerize :state, in: STATES, default: :proposed
 
@@ -25,7 +26,10 @@ class Event < ActiveRecord::Base
     [title, date].join(" : ")
   end
 
-  # To change the behaviour in /admin/events/1/edit...
+  def date
+    starts_at.strftime("%B %d")
+  end
+
   # driven by views/fields/enum_field/_show.html.erb
   def to_s
     title_with_date
--- a/db/migrate/20140929233719_create_events.rb
+++ b/db/migrate/20140929233719_create_events.rb
@@ -1,4 +1,8 @@
 class CreateEvents < ActiveRecord::Migration
+  class Event < ActiveRecord::Base
+    translates :title, :introduction, :conclusion
+  end
+
   def up
     create_table :events do |t|
       t.string :type, null: false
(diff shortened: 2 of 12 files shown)


app/models/event.rb
```diff
@@ -27,10 +28,14 @@ def location
   end
 
   def title_with_date
-    date = starts_at.strftime("%B %d")
+    return nil if starts_at.blank?
```
itsWill: Added this check because when an event is created in the admin with a empty `starts_at` field administrate calls this method before validation, which produces a no method for nil class error.

spec/views/envents/index.html.erb_spec.rb
```diff
@@ -0,0 +1,21 @@
+require "rails_helper"
```
benichu: This folder should be named: `spec/views/events` not `spec/views/envents`

app/models/event.rb
nicholasjhenry: @itsWill slightly more idiomatic Ruby to just do a return -- `nil` would be implied there. Example:

```
return if starts_at.blank?
```

itsWill:
rebased and ready to merge

nicholasjhenry:
Hey @itsWill -- thanks for putting this together. I wanted to get this merged in tonight but there's an error when running the migration from the start:

```
$ bin/rake db:migrate:reset
== 20140929233719 CreateEvents: migrating =====================================
-- create_table(:events)
   -> 0.0161s
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

Missing translated field :introduction/Users/nicholas/.rvm/gems/ruby-2.2.3@montreal-rb/gems/globalize-5.0.1/lib/globalize/active_record/migration.rb:139:in `b
lock in validate_translated_fields'
```

I know what the issue is, but a little tired now to think through a solution. Perhaps yourself or another contributor could take a look. Thanks!

db/migrate/20140929233719_create_events.rb
```diff
@@ -1,4 +1,8 @@
 class CreateEvents < ActiveRecord::Migration
+  class Event < ActiveRecord::Base
+    translates :title, :introduction, :conclusion
+  end
```
itsWill: The globalize translations are introduced in this migration then later removed, this causes an error in the  `rake db:migrate` .  This is a workaround to the issue @nicholasjhenry reported.

itsWill:
I added a work around to it, I managed to get it to work on my machine however without it, by doing a `rake db:reset` before the migrate.

itsWill:
rebased, and ready to merge.

sophiedeziel:
@itsWill can you rebase again? :)

sophiedeziel:
new -> #196 (that way I could commit on the branch)

PR #139, opened 2015-12-06
[chore] Adds rubocop to run ruby style checking locally and automatically

benichu:
Running automatically on file changes when you use `guard` in your development
process

cc @cawel

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,0 +1,8 @@
+inherit_from: .ruby-style.yml
+
+# Pending decision @ https://github.com/montrealrb/Montreal.rb/issues/131
+Style/StringLiterals:
+  Enabled: false
+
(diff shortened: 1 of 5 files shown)


cawel:
Much appreciated! :+1:  This will reduce the noise (generated by HoundCI) on issues, as people will be able to fix the code style issues *before* pushing.

@benichu Note that there's a rails option we could leverage in that `.rubocop.yml` file.

benichu:
Indeed, I set it up in guard: https://github.com/montrealrb/Montreal.rb/pull/139/files#diff-98f0657221f21d4a8adae4120e31c29bR28
Forgot it could be done in the config file :( can you add it when you have a chance?

cawel:
@benichu Will do. Within a few days. And I'll copy you.

Issue #114, opened 2015-12-04
Remove unused admin controllers/dashboards


nicholasjhenry:
Unfortunately we ended up with some generated code that we don't need for the admin (controllers/dashboard):

* companies
* meetups
* talks (future)
* users (future)
* votes

Some of these will be used in the future, but since they haven't been implemented fully we should remove them now. 

Note: These were generated by executing

```
rails g administrate:dashboard
```

rather than just generating the dashboard for the specific model:

```
rails g administrate:dashboard [model]
```

PR #148, opened 2015-12-09
Added talks dashboard

karimmtarek:
- Created talk model
- Added association with events
- Created talks dashboard
- Removed useless generated dashboards (see #114)
- Added factories for: talk, event, location
- Created events index view placeholder


spec/factories/talks.rb
```diff
@@ -0,0 +1,10 @@
+FactoryGirl.define do
+  factory :talk do
+    title "MyString"
```
sophiedeziel: We should use Faker for the factories instead of hardcoded values

karimmtarek: Noted :+1:

sophiedeziel:
The pull request title confuses me, is it still WIP or I can merge?

karimmtarek:
@sophiedeziel it still needs some work on the views side, as I'm not sure how are we planing to display talks, is it going to be displayed with each event, so in this case we don't need any views for the talks, or you can view talks individually so I should create and index and a show view for talks.
What do you think?

cawel:
@karimmtarek Imo, it does not make sense that both "requires-code-review" and "ready" labels be applied simultaneously. 

In addition to that, I think a PR without any label could be seen as _implicitly_ "ready", without the need for a label. And if some other PR gets merged into `master`, then that PR with the "ready" label would not be "ready" (to merge into `master`) anymore, because it would require a rebase on `master`.

karimmtarek:
@cawel you are totally right :+1:
Adding the "ready" label was actually a mistake, I was trying to use waffle.io and when I moved the task to the ready column I thought it was just a matter of a visual presentation, I wasn't ware it's going to add a label.
"ready" label removed.

cawel:
A side-effect from Waffle.io then ! No worries :)

PR #150, opened 2015-12-09
Fix #115 - add 'user_id' to 3 models: Event, Job, NewsItem

### LINKED ISSUE #115: Add `published_by` to events, jobs, news
This would assign the current user to these models

cawel:
- Add an `author` relationship to all of those 3 models
- Make the `user_id` field mandatory in both db and model
- Assign the `user_id` field to the current user when creating/updating any of those 3 models in the admin interface
- Empower `Faker` in a few factories
- Add model specs
- Add admin controller specs involving `Devise`
- Add a rake task to create a default user in the database
- Require the existence of the default user in db for the rake task importing the legacy db data
- Add methods related to the default user in the `User` model
- Remove useless `news_item_helper` and its spec
- Remove useless `home_helper`
- Fix old style rails code in model

[Review on Reviewable](https://reviewable.io/reviews/montrealrb/montreal.rb/150)

### DIFF
--- a/app/controllers/admin/events_controller.rb
+++ b/app/controllers/admin/events_controller.rb
@@ -1,19 +1,5 @@
 module Admin
   class EventsController < Admin::ApplicationController
-    # To customize the behavior of this controller,
-    # simply overwrite any of the RESTful actions. For example:
-    #
-    # def index
-    #   super
-    #   @resources = Event.all.paginate(10, params[:page])
-    # end
-
-    # Define a custom finder by overriding the `find_resource` method:
-    # def find_resource(param)
-    #   Event.find_by!(slug: param)
-    # end
-
-    # See https://administrate-docs.herokuapp.com/customizing_controller_actions
-    # for more information
+    include AdministrateCustomization
   end
 end
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -3,6 +3,7 @@ class Job < ActiveRecord::Base
   STATES = %w(draft published archived).freeze
 
   belongs_to :organization
+  belongs_to :author, foreign_key: :user_id, class_name: "User"
 
   scope :published, -> { where(state: :published).order(created_at: :desc) }
 
@@ -17,4 +18,5 @@ class Job < ActiveRecord::Base
   validates :state,
             presence: true,
             inclusion: { in: STATES }
+  validates :author, presence: true
 end
--- a/app/controllers/admin/jobs_controller.rb
+++ b/app/controllers/admin/jobs_controller.rb
@@ -1,4 +1,5 @@
 module Admin
   class JobsController < Admin::ApplicationController
+    include AdministrateCustomization
   end
 end
--- a/app/controllers/admin/news_items_controller.rb
+++ b/app/controllers/admin/news_items_controller.rb
@@ -1,19 +1,5 @@
 module Admin
   class NewsItemsController < Admin::ApplicationController
-    # To customize the behavior of this controller,
-    # simply overwrite any of the RESTful actions. For example:
-    #
-    # def index
-    #   super
-    #   @resources = NewsItem.all.paginate(10, params[:page])
-    # end
-
-    # Define a custom finder by overriding the `find_resource` method:
-    # def find_resource(param)
-    #   NewsItem.find_by!(slug: param)
-    # end
-
-    # See https://administrate-docs.herokuapp.com/customizing_controller_actions
-    # for more information
+    include AdministrateCustomization
   end
 end
--- a/app/controllers/concerns/admin/administrate_customization.rb
+++ b/app/controllers/concerns/admin/administrate_customization.rb
@@ -1,0 +1,11 @@
+module Admin
+  module AdministrateCustomization
+    extend ActiveSupport::Concern
+
+    private
+
+    def resource_params
+      super.merge(user_id: current_user.id)
+    end
+  end
+end
--- a/app/models/event.rb
+++ b/app/models/event.rb
@@ -12,12 +12,14 @@
 class Event < ActiveRecord::Base
   translates :title, :introduction, :conclusion
   belongs_to :location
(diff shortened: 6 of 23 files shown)


cawel:
Rebased on top of `master`!

And only after the rebase do I have a "coverage decreased (-0.2%)", so I'm calling treason on that one ;)

cawel:
Rebased again!

nicholasjhenry:
@cawel Thanks for putting this together. Can you please check to see if this causes any problems with the importing of the old posts? May need a default user for the import (see the rake task).

nicholasjhenry:
@cawel: @benichu informs me the task has already been run. So we'll need a user for the import default news items please.

cawel:
Indeed, otherwise I get:

```ruby
ActiveRecord::RecordInvalid: Validation failed: Published by can't be blank
```

So you want me to modify the migration which adds a `user_id` column into the 'news_items' table so that it _also_ sets a value for that new column corresponding to the id of a default user (which I will also have to create in the 'users' table)?

nicholasjhenry:
I think that would be the best way to handle it -- user should be deactivated.

cawel:
Roger that @nicholasjhenry. I should have that ready by the end of tomorrow.

cawel:
@nicholasjhenry Since both `confirmable` and `lockable` are not available options in our current devise configuration, what did you have in mind when you said "user should be deactivated"?

Maybe [overriding Devise's `active_for_authentication?` method](https://github.com/plataformatec/devise/blob/master/lib/devise/models/authenticatable.rb#L35-L45) in the `User` model, like so?

```ruby
def active_for_authentication?
  super && email != 'default.user@domain.com'
end
```

nicholasjhenry:
@cawel I am fine with whatever you think is appropriate.

cawel:
@nicholasjhenry @benichu :

* I modified the `database:legacy:import_news` rake task so that it requires the default user to be present in the database. 
* I added the same kind of requirement in the migration which adds the `user_id` column to the 'news_items' table, since I wanted to make that `user_id` column not nullable.
* I made sure the default user cannot authenticate, using the `active_for_authentication?` method trick :boom: as described in [my previous comment](https://github.com/montrealrb/Montreal.rb/pull/150#issuecomment-163822568).
* I rebased on top of `master`.

That's a big PR. A code review would not hurt ;)

benichu:
@cawel Thanks a lot for that PR, I will review it very soon (but most probably next week)
We are currently focusing on making sure we have a "shipable" website for next tuesday's montreal.rb meetup. That is why some of the PR are not going to be merged by then...

I'll also do my best to update the roadmap based on what we have and where we could realistically be next year... This will be a good foundation for properly shipping new interesting stuff, organizing focused workshops and efficiently using every contributors skills and time without burning them out or the project itself :)

cawel:
@benichu Wise planning :+1:

app/models/job.rb
```diff
@@ -3,4 +3,7 @@ class Job < ActiveRecord::Base
   STATUSES = [:draft, :published, :archived]
   enumerize :state, in: STATUSES
   belongs_to :organization
+  belongs_to :published_by, foreign_key: :user_id, class_name: "User"
```
karimmtarek: Here is a rookie question :)   Why not do it like this `belongs_to :user`?
I'm sure there is a good reason for doing it like this, I just wanna be educated, that's all.

cawel: It could totally be done the way you wrote: `belongs_to :user`. It would have the benefit of being terser.

However, I would invoke the argument of "code readability" as a justification here. I think the concept of a "publisher" describes the problem domain better than a "user". When attempting to find out who published (i.e. authored) the model (be it a job, an event or a news_item), I think

```ruby
Job.first.published_by
```

reads better than:

```ruby
Job.first.user
```

As an alternative, we could define that `belongs_to` relationship with `publisher` (or even `author`) instead of `published_by`.

karimmtarek: Thanks for explaining, your argument makes total sense :+1: 

> As an alternative, we could define that `belongs_to` relationship with `publisher` (or even `author`) instead of `published_by`.

I'm leaning towards using `author` as any created item (be it a job, an event or a news_item) doesn't have to end up being published for whatever reason, if not then `publisher` would be my second pick :)

cawel: Compelling argument. Now those 3 models `belongs_to` an 'author' instead of 'published_by'. 

Changes done ! :boom:

karimmtarek: Perfect :+1:

steakunderscore:
I rebased this patch and put it up for review. Feel free to claim it back @cawel. Not sure why github made a new PR. But it's at https://github.com/montrealrb/Montreal.rb/pull/185.

cawel:
@steakunderscore Thanks for the ping. I do reclaim that PR then :)

I rebased that very PR on `master`, potentially rendering #185 obsolete (which needed a rebase on master anyways).

sophiedeziel:
Yay!!! I'm reviewing, thanks @cawel and @steakunderscore

app/controllers/admin/events_controller.rb
```diff
@@ -15,5 +15,11 @@ class EventsController < Admin::ApplicationController
 
     # See https://administrate-docs.herokuapp.com/customizing_controller_actions
     # for more information
+
+    private
+
+    def resource_params
```
sophiedeziel: That method is duplicated in three controllers. I think it would be a good idea to create a concern (or a good ol' ruby module) to include. If we want to modify it (if we want to let the admin override the value to write on behalf of another user for example), we'll do it once.

cawel: Good catch. I created a `Admin::AdministrateCustomization` concern (and included it in those 3 controllers) and tested it in the admin interface and it works. Cool?

cawel:
Freshly rebased on `master`!

Issue #153, opened 2015-12-11
Footer links are not linked to anything


benichu:

karimmtarek:
I can fix that, just need the missing links
- https://twitter.com/montrealrb
- facebook?
- google+?

PR #156, opened 2015-12-11
WIP: Add footer links

karimmtarek:
for #153
I'm not sure if the G+ link is meant to be the google group or we have another url for that, also I couldn't find  the facebook url so I commented it for now until I get the url or just remove it.
Should those links open in a new tab? Should I add 'no-follow' to them?

[Review on Reviewable](https://reviewable.io/reviews/montrealrb/montreal.rb/156)


benichu:
Let's go with those 2 links for now. Whenever you are ready I can merge this.

karimmtarek:
I think it's ready, thanks.

PR #164, opened 2015-12-13
Pages: empower the 'friendly_id' gem to generate slugs

### LINKED ISSUE #35: Manage and display pages
We need pages of content (separated from the news feed) for special content (about, FAQs, howtos)

- [x] manage pages in the admin
- [ ] access pages via a link

cawel:
- Slug history is enabled (by creating a new db table)
- Slug is updated if the `title` attribute changes
- Show `slug` field on the index page
- Show `slug` field on the show page
- Fix #35

### DIFF
--- a/db/migrate/20151213185304_add_slug_to_page.rb
+++ b/db/migrate/20151213185304_add_slug_to_page.rb
@@ -1,0 +1,12 @@
+class AddSlugToPage < ActiveRecord::Migration
+  def change
+    add_column :pages, :slug, :string, null: false, default: 'temporary-slug'
+
+    # make sure we define a slug for existing pages
+    Page.find_each do |page|
+      page.update(slug: '')
+    end
+
+    add_index :pages, :slug, unique: true
+  end
+end
--- a/Gemfile
+++ b/Gemfile
@@ -1,8 +1,6 @@
 source "https://rubygems.org"
 ruby "2.2.4"
 
-# Bundle edge Rails instead: gem 'rails', github: 'rails/rails'
-gem "rails", "~> 4.2.2"
 
 # Database
 gem "pg"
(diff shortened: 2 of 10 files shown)


mathieugagne:
:+1: Good work!

cawel:
Freshly rebased on `master`!

db/migrate/20151213185304_add_slug_to_page.rb
```diff
@@ -0,0 +1,6 @@
+class AddSlugToPage < ActiveRecord::Migration
+  def change
+    add_column :pages, :slug, :string, null: false
```
sophiedeziel: When the database already contains pages, that error is raised:
```
== 20151213185304 AddSlugToPage: migrating ====================================
-- add_column(:pages, :slug, :string, {:null=>false})
rake aborted!
StandardError: An error has occurred, this and all later migrations canceled:

PG::NotNullViolation: ERROR:  column "slug" contains null values
```
It's tricky because we can't put a default value...

cawel: Yeah, that's exactly what I'm trying to fix at this very moment. Didn't have the time to fix it before someone noticed it ;)

cawel: Here is what I propose (as can be read in the `db/migrate/20151213185304_add_slug_to_page.rb` file):

```
add_column :pages, :slug, :string, null: false, default: 'temporary-slug'

# NOTE: we'll need to add a unique index on that 'slug' column after:
# 1) the current migration has been run in production
# 2) all slugs from pages with a 'temporary-slug' slug have been changed

# The unique index on the 'slug' column should look like this:
# add_index :pages, :slug, unique: true
```

sophiedeziel:
About that migration...

I did some tests on my end:

I changed the implementation of `should_generate_new_friendly_id?` in the page model to have :
```
  def should_generate_new_friendly_id?
    slug.blank? || changed.include?("title")
  end
```

Then, in the rails console, I did
```
Page.find_each do |page|
  page.update_attribute(:slug, nil)
end
```
Unique slugs for everyone!!!

cawel:
@sophiedeziel :+1: 

Then here is the new version of the migration:

```ruby
def change
  add_column :pages, :slug, :string, null: false, default: 'temporary-slug'

  # make sure we define a slug for existing pages
  Page.find_each do |page|
    page.update(slug: '')
  end
  add_index :pages, :slug, unique: true
end
```

sophiedeziel:
Fantastic!!

PR #175, opened 2016-02-22
Better factories with linting and traits

steakunderscore:
* Wrote the rest of the factories
+ Added test for factory lint
+ Added seeds to get developers started faster
* Fixed tests to use factory traits, rather than separate factories

### DIFF
--- a/db/seeds.rb
+++ b/db/seeds.rb
@@ -1,0 +1,16 @@
+require "factory_girl_rails"
+
+FactoryGirl.create(:user, email: "user@example.com", password: "password")
+FactoryGirl.create(:user, :admin, email: "admin@example.com", password: "password")
+
+FactoryGirl.create(:job, :published)
+FactoryGirl.create(:news_item, :published)
+FactoryGirl.create(:news_item, :slugged)
+
+FactoryGirl.create_list(:event, rand(5..20))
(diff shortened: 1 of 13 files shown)


mathieugagne:
I'm good with this PR. :+1: 
Nice work with the factory traits.

@cawel do you second?

mathieugagne:
Why would coveralls complain if coverage didn't decreased? I'm not a big user of coveralls. Anyone got an explanation?

cawel:
@mathieugagne All those spec improvements look good to me. Nice work indeed :+1: 

As for seeding using FactoryGirl, [ThoughtBot actually discourages this practice](https://robots.thoughtbot.com/factory_girl-for-seed-data), but I remain unconvinced by their argument. And I have done that in the past several times (seeding using FactoryGirl) and so far I haven't been disappointed.

As for the coveralls, I'm also new to that tool. I can see [on the build report page](https://coveralls.io/builds/5163126) that the files in the "changed" list do not appear to have their coverage changed. Moreover, those "changed" files are specs, where coverage should be irrelevant. I'm clueless.

mathieugagne:
FactoryGirl in seeds. :ok: 
Coveralls. :ok: :confused: Will pass.

PR #177, opened 2016-02-25
Fix issues/93 job validation

steakunderscore:
This is the patch written by @AndyObtiva rebased.

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -1,6 +1,19 @@
 class Job < ActiveRecord::Base
   extend Enumerize
-  STATUSES = [:draft, :published, :archived]
-  enumerize :state, in: STATUSES
+
+  STATES = %w(draft published archived).freeze
+
   belongs_to :organization
+
+  enumerize :state, in: STATES
+
+  validates :title,
+            presence: true,
+            length: { maximum: MAX_STRING_COLUMN_LENGTH }
+  validates :description,
+            presence: true,
+            length: { maximum: MAX_TEXT_COLUMN_LENGTH }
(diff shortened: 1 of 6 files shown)


app/models/job.rb
```diff
@@ -1,6 +1,15 @@
 class Job < ActiveRecord::Base
   extend Enumerize
-  STATUSES = [:draft, :published, :archived]
-  enumerize :state, in: STATUSES
+  STATES = %w(draft published archived)
+  enumerize :state, in: STATES
   belongs_to :organization
+  validates :title,
+            presence: true,
+            length: { maximum: MAX_STRING_COLUMN_LENGTH }
```
steakunderscore: What are peoples thoughts on having lengths? I don't normally have them, PG and all..

mathieugagne: I usually don't do it either but I find it'd be good practice for Rails to validate that it won't blow up at the db level. I'd rather have an error message telling me to reduce my text size than a 500 server error cause PG complained.

cawel: I second.

app/models/job.rb
```diff
@@ -1,6 +1,19 @@
 class Job < ActiveRecord::Base
   extend Enumerize
-  STATUSES = [:draft, :published, :archived]
-  enumerize :state, in: STATUSES
+
+  STATES = %w(draft published archived).freeze
```
steakunderscore: This should be symbols not strings. So should be using `%i` not `%w`.

mathieugagne:
All good! :+1:

PR #180, opened 2016-02-28
[issue#9] Display job listings

steakunderscore:
+Added Job controller with index and show
+Added link to jobs page to header
~Formatted code in job model
*Fixed draft job trait name

### DIFF
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -1,0 +1,18 @@
+class JobsController < ApplicationController
+  before_action :set_job, only: [:show]
+
+  # GET /jobs
+  def index
+    @jobs = Job.published
+  end
+
+  # GET /jobs/1
+  def show
+  end
+
+  private
+
+  def set_job
+    @job = Job.published.find(params[:id])
+  end
+end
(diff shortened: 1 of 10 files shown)


mathieugagne:
@steakunderscore branch has conflicts. Please rebase

app/helpers/jobs_helper.rb
```diff
@@ -0,0 +1,2 @@
+module JobsHelper
```
mathieugagne: Do we need this or can we delete it?

app/controllers/jobs_controller.rb
```diff
@@ -0,0 +1,18 @@
+class JobsController < ApplicationController
+  before_action :set_job, only: [:show, :edit, :update, :destroy]
```
mathieugagne: I can't see edit, update & destroy here. Perhaps might even move the before_action back to #show

app/helpers/jobs_helper.rb
steakunderscore: I left it there because there are other empty helpers left in `app/helpers/`. I might just put up a patch removing them all. See PRl#183.

mathieugagne: :+1: Good thinking

steakunderscore:
Ditching this PR in favour of 183

PR #184, opened 2016-03-03
Fix mistake in event factory

steakunderscore:
The faker `between` methods can't handle `ActiveSupport::Duration`
objects. It's expecting something that can responds to `:to_date`. So
needed to append `.from_now` to adhere to this.


mathieugagne:
:+1: Good catch
I can't merge on my phone. Will do later

PR #185, opened 2016-03-03
User id to some models

steakunderscore:
Starting to rebase other peoples commits.
I'm actually not a big fan of this commit. It's far too big and could easily be split up. What do others think?


steakunderscore:
I have split a few things out now, it would be good to get a code review .

sophiedeziel:
The original pull request #150 is the "official" one.

Issue #105, opened 2015-12-04
Add Event#state


nicholasjhenry:
This allows to propose events for the future without publishing them on the site.

- state values are: (proposed, scheduled)
- consider how this affects display events on the home page body, home page sidebar, and event index.

itsWill:
I'll pick this one up as well since I touch it when I work on https://github.com/montrealrb/Montreal.rb/issues/48

itsWill:
will wait till https://github.com/montrealrb/Montreal.rb/issues/128 is resolved to finish up this one

PR #192, opened 2016-03-08
[issue #105] add state to event

sophiedeziel:
This solves #105


benichu:
rebase and good to go.

Issue #160, opened 2015-12-12
It would be nice to render markdown fields in the admin


benichu:
<img width="947" alt="screenshot 2015-12-04 11 47 55" src="https://cloud.githubusercontent.com/assets/40253/11595340/e2349988-9a7c-11e5-95f6-de09362f0c06.png">

benichu:
follow-up to #83

PR #193, opened 2016-03-08
Created a special field to render markdown in administrate

sophiedeziel:
This solves the issue #160


spec/factories/talks.rb
```diff
@@ -0,0 +1,10 @@
+FactoryGirl.define do
+  factory :talk do
+    title { Faker::Book.title }
+    description { Faker::Lorem.sentence }
+    state { Talk::STATES.sample }
+    level "beginner"
+    slides_url { Faker::Internet.url }
+    event
```
steakunderscore: I think rather than just `event`, perhaps `association :event` would be better?

spec/models/event_spec.rb
```diff
@@ -23,5 +36,11 @@
       event = create(:event, author: author)
       expect(event.author).to eq author
     end
+
+    it "does not validate when 'body' is blank" do
+      event = Event.new(title: nil)
```
steakunderscore: I might have missed the point here, but should that be `body: nil` not `title: nil`?

spec/controllers/admin/events_controller_spec.rb
```diff
@@ -25,7 +25,7 @@
   end
 
   describe "PUT #update" do
-    let(:event) { create :event }
+    let!(:event) { FactoryGirl.create(:event) }
```
sophiedeziel: It's an old habit, thanks for mentioning! I'll fix it.

PR #195, opened 2016-03-09
update rails from 4.2.5.2 to 4.2.6

sophiedeziel:


Gemfile
```diff
@@ -20,7 +20,7 @@ gem 'uglifier', '>= 1.3.0'
 gem 'sprockets', '~> 3.0'
 
 # Application
-gem 'rails', '~> 4.2.2'
+gem 'rails', '~> 4.2'
```
steakunderscore: Hmm.. Is this what you want? I personally would have not changed this line, so maybe just double check this is correct.

sophiedeziel: I don't want to fix the minor version because there is no reason to do so. I prefer more flexible dependencies.

PR #201, opened 2016-03-17
Minor changes

steakunderscore:


db/schema.rb
```diff
@@ -138,6 +138,8 @@
     t.string   "slug",       default: "temporary-slug", null: false
   end
 
+  add_index "pages", ["slug"], name: "index_pages_on_slug", unique: true, using: :btree
```
sophiedeziel: Was that added automatically? Is it in a migration?

sophiedeziel: I just checked, yes it's in a migration :+1:

PR #216, opened 2016-04-06
Update README.md

bluciam:
The official site url's protocol is not https, wiith it there is an error. Changed to http protocol.


sophiedeziel:
Thank you!!

PR #219, opened 2016-04-08
Side bar specific styling

kiramclean:


app/views/news_items/index.html.erb
```diff
@@ -1,4 +1,4 @@
 <% provide :title, 'News Items' %>
 
 <h2>News</h2>
-<%= render partial: "news_items/news_item", collection: @news_items %>
```
sophiedeziel: Can you revert that change so the login doesn't break?

sophiedeziel:
I merged it manually, fixing some minor details...

PR #222, opened 2016-04-23
fix GitHub authentication

### LINKED ISSUE #154: Signing in with github does not seem to be working, any missing configurations?

sophiedeziel:
Fixes #154

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -28,8 +28,8 @@ class User < ActiveRecord::Base
   has_one :member
 
   def self.from_omniauth(auth)
-    where(provider: auth.provider, uid: auth.uid).first_or_create do |user|
-      user.email = auth.info.email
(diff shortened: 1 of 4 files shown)


sophiedeziel:
@benichu That will fix your login issue. Want to review?

app/models/user.rb
```diff
@@ -27,29 +27,51 @@ class User < ActiveRecord::Base
...
+    end
 
-  def self.default_user
-    User.find_by(email: DEFAULT_USER_EMAIL)
-  end
+    def default_user
+      User.find_by(email: DEFAULT_USER_EMAIL)
+    end
+
+    private
```
steakunderscore: Can't have private class methods..?

sophiedeziel: They actually are private class methods because they are inside of `class << self`

PR #223, opened 2016-04-29
Auto-tweet when an event is published

sophiedeziel:


itsWill:
LGTM I would :shipit: just to have the testing/twitter infrastructure in place. However I think the next step is to refactor `TweetEvenService` into a general `TweetMessageService` that isn't bound to a specific model and we can reuse with new items, job postings, etc..

sophiedeziel:
Yeah, I am not yet satisfied with that pull request. There is still some polish I want to do. I didn't want to extract abstraction prematurely, but I guess that one won't do any harm.

PR #224, opened 2016-05-01
#203 Added admin checkbox to user dashboard

### LINKED ISSUE #203: An admin should be able to set another user as an admin

steakunderscore:
This should close #203


sophiedeziel:
Thanks!!!

PR #225, opened 2016-05-12
Added Emoji Support :clap: :fire: :clap:

itsWill:
💥 Emojis are here! 💥  

I've added some tests to keep it 💯 

This is implemented using the ⚡  `gemoji` gem ⚡ 

I've also made all titles titleized but don't know if that is always appropriate ❓ ❔ ❓ 

☀️  emojineers at 🚜  ☀️

PR #226, opened 2016-05-16
Added Emoji Support :clap: :fire: :clap: #225

itsWill:
💥 Emojis are here! 💥

I've added some tests to keep it 💯

This is implemented using the ⚡ gemoji gem ⚡

I've also made all titles titleized but don't know if that is always appropriate ❓ ❔ ❓

☀️ emojineers at 🚜 ☀️

New PR to fix 📛  tests


app/helpers/emoji_helper.rb
```diff
@@ -0,0 +1,12 @@
+module EmojiHelper
+  def emojify(content)
+    h(content).to_str.gsub(/:([\w+\s-]+):/) do |match|
+      name = $1.tr(" ", "_").downcase
+      if emoji = Emoji.find_by_alias(name)
+        %(<img alt="#{name}" src="#{image_path("emoji/#{emoji.image_filename}")}" style="vertical-align:middle" width="20" height="20" />)
```
sophiedeziel: A css class is prefered. Remove the inline style. Also, width and height should be set in `em` to scale with font-size.

app/views/events/_event.html.erb
```diff
@@ -1,11 +1,11 @@
 <div class="card">
-  <h1><%= link_to event.title.titleize, event %></h1>
+  <h1><%= link_to emojify(event.title.titleize), event %></h1>
   <div class="event-meta">
     <p><%= icon 'calendar-o', class: 'fa-fw' %><%= event.starts_at.strftime('%B %e, %Y  -  %k:%M') %></p>
     <p><%= icon 'map-marker', class: 'fa-fw' %><%= event.location.name %>, <a href="https://www.google.ca/maps/search/<%= event.location.address %>", target='_blank'><%= event.location.address %></a></p>
   </div>
   <div class="markdown-body">
-    <%= render_markdown_as_html event.body %>
+    <%= render_markdown_as_html emojify(event.body) %>
```
sophiedeziel: The emojify helper should wrap `render_markdown_as_html`, not the opposite. Emojify will generate html tags in a text formatted with markdown.

app/helpers/emoji_helper.rb
```diff
@@ -0,0 +1,12 @@
+module EmojiHelper
+  def emojify(content)
```
sophiedeziel: Add some rspec specs to test that feature.

sophiedeziel:
There's a bunch of tools you can use before commiting or pushing. `rubocop -a` will auto-fix most things hound complains about. I love it 😸

app/views/events/_event.html.erb
itsWill: I actually think makes sense to include the `emojify` helper within the `render_markdown_as_html` helper for two reasons:  

1. The way I see it github emoji i.e colon emoji colon syntax is part of the github markdown so it should be treated as such.
2.  Calling the `emojify` method shouldn't result in markdown being outputted, only the replacement of the :emoji: syntax with the right emoji, that doesn't seem like expected behavior.

 I see the `emojify`method as a helper method to the `render_markdown_as_html` method which replaces "markdown emoji syntax" with the right emoji.

sophiedeziel: Yes!

itsWill:
Refactored to add `emojify` method as part of the `render_markdown_as_html` method.

 Added `byebug` to the `Gemfile` for those of us who use it instead of `pry`. I've had to update the `.gitignore` as well.

PR #230, opened 2016-05-26
Set the right ruby version in README

dechuck:
The Gemfile is now using 2.3.1.


sophiedeziel:
Hahaha!! Thank you for pointing it out!!!!

dechuck:
😄

PR #248, opened 2017-01-18
Fix rubocop errors and migrations for SQLite support in development/test env

caristomenopoulos:


steakunderscore:
Running sqlite in dev is not something I ever do. I much prefer to run postgres, even if it's in a docker container. But there is the benefit of having sqlite to lower the bar for others to help out.

The rubocop fixes are good.

Issue #242, opened 2016-11-23
Send automatically an invite to our Slack community on sign up


sophiedeziel:
The idea is to get rid of https://montrealrb-slack.herokuapp.com/. On sign up on the website, an invite should be automatically sent, using the Slack API to do so.

PR #251, opened 2017-03-03
Auto-email Montreal.rb slack invite to new users

marina101:
(work-in-progress, needs admins to try it out since I'm not a slack admin for the team, so can't generate the auth code)

Addresses #242. When a new user signs up, this takes the user's email and sets off a flow to get slack to auto-send them an invite email for the Montreal.rb slack team. The code for the script is inspired by https://ruby.unicorn.tv/screencasts/automatically-send-slack-invitations. Interestingly, it doesn't seem that the slack api has an official endpoint for this, but there are a few scripts that use this unofficial endpoint.

Someone who is an admin for the slack team should try this out in development, specifying the slack channel id to invite the user too, as well as the auth token (which should be and ENV variable and shouldn't be committed to version control). Though it would be great to have a different test auth token so that a test can be written for this as well.
@sophiedeziel @garyharan

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,4 +1,7 @@
 # frozen_string_literal: true
+require "open-uri"
+require "colorize"
+
 module ApplicationHelper
   def full_title(page_title = "")
     base_title = "Montreal.rb"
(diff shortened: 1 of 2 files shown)


sophiedeziel:
Thank you for that pull request! It is a very nice addition to the project.

I will test it locally pretty soon. I also have some suggestions to improve the code that we can discuss. I am about to do a proper review.

app/helpers/application_helper.rb
```diff
@@ -8,4 +11,32 @@ def full_title(page_title = "")
       page_title + " | " + base_title
     end
   end
+
+  def email_slack_invite(user_email)
```
sophiedeziel: This is the first time I see a helper used for something else than extracting logic from the views. We usually use service classes (there are already some examples in `app/services`).

app/helpers/application_helper.rb
```diff
@@ -8,4 +11,32 @@ def full_title(page_title = "")
...
+    # Get the channel ID from https://slack.com/api/channels.list?token=<YOUR_TOKEN>
+    channels = ""
```
sophiedeziel: You can introduce environment variables here.

app/helpers/application_helper.rb
```diff
@@ -8,4 +11,32 @@ def full_title(page_title = "")
...
+    url += "&email=#{URI.escape(user_email)}"
+
+    # Make the request and capture the response
+    response = JSON.parse(open(url).read)
+
+    if response["ok"]
+      Rails.logger.info("slack invitation sent to: #{user_email}".colorize(:green))
+    else
+      Rails.logger.info("slack invitation failed to send to #{user_email}".colorize(:red))
+      Rails.logger.info(response.inspect.colorize(:yellow))
```
sophiedeziel: That's a very good idea to use the Rails logger here. I like it!

app/helpers/application_helper.rb
```diff
@@ -8,4 +11,32 @@ def full_title(page_title = "")
...
+
+    # Construct the API request URL
+    # This is not documented by Slack, and may not work forever
+    # As it is an unofficial API endpoint, use at your own risk!
+    url = "https://slack.com/api/users.admin.invite"
+    url += "?channels=#{channels}"
+    url += "&set_active=true"
+    url += "&_attempts=1"
+    url += "&token=#{token}"
+    url += "&email=#{URI.escape(user_email)}"
```
sophiedeziel: Using [`HTTP::URI.build`](http://ruby-doc.org/stdlib-2.0.0/libdoc/uri/rdoc/URI/HTTP.html#method-c-build) instead of concatenating a string could lead to more readable and maintainable code.

sophiedeziel:
Closing because #253

PR #268, opened 2017-05-13
Feature/searchable jobs

bellef:
Hey, it's my first pull request on this project (and my first contribution to an open source project).
I though it would be great if we could search through jobs in the index page.

It looks like that :
selection_140


app/controllers/jobs_controller.rb
```diff
@@ -15,4 +21,8 @@ def show; end
   def set_job
     @job = Job.published.find(params[:id])
   end
+
+  def filter_jobs(query, base_relation)
+    base_relation.where("title LIKE ? OR description LIKE ?", "%#{query}%", "%#{query}%")
```
xymbol: Most probably, this shouldn't be here. Finding jobs matches is a model concern.

app/views/jobs/index.html.erb
```diff
@@ -2,4 +2,13 @@
 
 <h1>Listing Jobs</h1>
 
+<%= form_tag("/jobs", method: "get") do %>
```
xymbol: The Rails way is to always stick to url helper methods in views. In this case, `jobs_path`.

spec/controllers/jobs_controller_spec.rb
```diff
@@ -30,6 +33,24 @@
...
       expect(response).to render_template :index
     end
+
+    context "with a search query" do
+      before :each do
+        get :index, q: "foobar"
+      end
+
+      it "includes job when the query matches its title" do
+        expect(assigns(:jobs)).to include matching_title_job
```
xymbol: Moving this logic to the model makes testing more straightforward too. Tests here should probably exercise passing the optional search param only.

When setting up test data it's good practice to make them related to the problem domain. "Foo" and "bar" are okay if, for example, you're exercising a feature on plain strings. Here "backend developer" or "Québec" would make more sense as jobs titles.

bellef:
Thank you for your feedback @xymbol , I've tried to do better. ;)

xymbol:
@bellef Hound CI runs [RuboCop](https://github.com/bbatsov/rubocop) for Ruby which you can also use locally.

See: https://houndci.com/configuration.

spec/models/job_spec.rb
```diff
@@ -54,4 +54,30 @@
       expect(job.author).to eq author
     end
   end
+
+  describe ".search" do
+    it "returns a job when the query matches its title" do
+      matching_title_job = create(:job, :published, title: "Fullstack developer")
+      expect(Job.search("fullstack")).to include matching_title_job
+    end
```
sophiedeziel: I know the specs for that class are not very consistent. But in an effort to bring more clarity to those, I suggest that you use `subject` and `let` for your tests instead of declaring variables and call the method directly in the expectation.

sophiedeziel:
@bellef I love Rubocop's option `-a` which fixes automatically some kinds of "violations". Like alignment, simple-double-quotes and indentation. Then, you're left with a small number of fixes to do!

app/models/job.rb
```diff
@@ -20,4 +20,10 @@ class Job < ActiveRecord::Base
             presence: true,
             inclusion: { in: STATES }
   validates :author, presence: true
+
+  def self.search(search)
```
sophiedeziel: I forgot about that comment. 

You can use a named scope for that. Named scopes are pretty useful to chain ActiveRecord queries. You wont have to use a `base_relation` variable (but I like your usage of a guard clause).

spec/models/job_spec.rb
```diff
@@ -54,4 +54,43 @@
       expect(job.author).to eq author
     end
   end
+
+  describe ".search" do
+    let(:matching_description_job) do
```
sophiedeziel: You can shamelessly one-line those `let`!

spec/models/job_spec.rb
```diff
@@ -54,4 +54,43 @@
...
+      create(
+        :job,
+        :published,
+        title: "Fullstack developer",
+        description: "Familiarity with agile methodologies, especially Scrum or Kanban"
+      )
+    end
+
+    it "returns a job when the query matches its title" do
+      expect(subject.class.search("fullstack")).to include matching_title_job
```
sophiedeziel: `subject` can be defined the same way as `let`. 

Example:
```
subject { described_class.search('fullstack') }

it 'does something'
  expect(subject).to eq 5
end
```

Having the subject executing what you test allows the short implicit form. 

```
it { is_expected.to include matching_title_job }
```

😹

app/models/job.rb
xymbol: In most recent versions of Rails the preference for using a named scope is subject to personal preference. This is extracted from the [guides](http://guides.rubyonrails.org/active_record_querying.html#scopes):

>This is exactly the same as defining a class method, and which you use is a matter of personal preference.

As long as a model method returns an `ActiveRecord::Relation` it can be chained, etc. Scopes might be more readable but, methods are easier to extract. It's Rails versus Ruby.

This is the implementation for named scopes in `activerecord`. https://github.com/rails/rails/blob/010e246756c09f44e901f4fd8e8eab2cb3022e95/activerecord/lib/active_record/scoping/named.rb#L143

bellef:
One thing I still don't get: why does the tests coverage has decreased to 98%?

spec/models/job_spec.rb
```diff
@@ -54,4 +54,15 @@
       expect(job.author).to eq author
     end
   end
+
+  describe ".search" do
+    let(:matching_title_job) { create(:job, :published, title: "Fullstack developer") }
+    let(:matching_description_job) { create(:job, :published, description: "If you're a fullstack developer, this job is for you!") }
+    let(:not_matching_job) { create(:job, :published, title: "RoR developer", description: "Familiarity with agile methodologies, especially Scrum or Kanban") }
```
bellef: @sophiedeziel That's why it was on two lines: lines were too long.
What's best then: long lines or CI pleasure? :smile:

xymbol: @bellef Forget about the CI, long lines are bad for people.

You can use multi-line blocks with `let`.

```ruby
let(:non_matching_job) do
  create :job, :published,
         title: "RoR developer",
         description:
           "Familiarity with agile methodologies, especially Scrum or Kanban"
end
```

app/models/job.rb
```diff
@@ -7,6 +7,9 @@ class Job < ActiveRecord::Base
   belongs_to :author, foreign_key: :user_id, class_name: "User"
 
   scope :published, -> { where(state: :published).order(created_at: :desc) }
+  scope :search, (lambda do |query|
```
xymbol: The "stabby lambda" form is [recommended](http://guides.rubyonrails.org/active_record_querying.html#scopes) for scopes. You can also avoid repetition using the [hash conditions](http://guides.rubyonrails.org/active_record_querying.html#hash-conditions) form.

```ruby
scope :search, -> (query) {
  where "title LIKE :pattern OR description LIKE :pattern",
    pattern: "%#{query}%"
}
```

The Query Interface guide is great at covering all of this. http://guides.rubyonrails.org/active_record_querying.html

spec/models/job_spec.rb
```diff
@@ -54,4 +54,15 @@
       expect(job.author).to eq author
     end
   end
+
+  describe ".search" do
+    let(:matching_title_job) { create(:job, :published, title: "Fullstack developer") }
+    let(:matching_description_job) { create(:job, :published, description: "If you're a fullstack developer, this job is for you!") }
```
sophiedeziel: I'll edit the params to allow longer lines. That's a rule I dislike a lot

spec/models/job_spec.rb
```diff
@@ -56,41 +56,13 @@
...
-    it "returns a job when the query matches its description" do
-      expect(subject.class.search("agile")).to include matching_description_job
-    end
-
-    it "does not return a job when the query does not match its title nor its description" do
-      expect(subject.class.search("remote")).not_to include not_matching_job
-    end
+    it { is_expected.to include matching_title_job }
+    it { is_expected.to include matching_description_job }
+    it { is_expected.not_to include not_matching_job }
```
sophiedeziel: Much better! I :heart: it

app/models/job.rb
bellef: Thank you, I definitely should read more doc!

bellef:
@xymbol Then what should I do?
selection_180

bellef:
I still don't get why do the specs fail with:
```
An HTTP request has been made that VCR does not know how to handle:
         POST https://api.twitter.com/oauth2/token
```

I didn't even touched this part of the code...
Sorry for bothering you guys

xymbol:
@bellef Re the auth related test failure, it can be related to usual problems with TDD:

- Interdependent tests: when test B fails or succeeds when A fails or succeeds
- Intermittent tests: when tests sometimes pass and sometimes fail

Re your changes, did you try running the suite with your changes applied? I did and encountered 2 valid tests failing and, confirmed one introduced bug. The dropped scope chain for jobs breaks pre-existing tests. It should be:

```ruby
  def index
    @jobs = Job.published.includes(:organization).search(params[:search])
  end
```

Apply this fix and see what Travis and Coveralls say.

bellef:
How about now @sophiedeziel @xymbol ? :)

bellef:
Thanks @sophiedeziel , now what should we do about this PR? Should I close it?

sophiedeziel:
The feature is good but I'm not done with syncing with master and reconfiguring the github repository. I disabled Hound because it's overwhelming!

I like the search feature and it will set grounds for search in other pages. I'll merge when I'll be ready.

app/models/job.rb
```diff
@@ -10,6 +10,13 @@ class Job < ActiveRecord::Base
   belongs_to :organization
 
   scope :published, -> { where(state: :published).order(created_at: :desc) }
+  scope :search, (lambda do |query|
+    if query.present?
```
sophiedeziel: You can use a guard clause here, like
```
scope :search, -> do
  return all unless query.present?
  where(...)
end
```

app/models/job.rb
```diff
@@ -10,6 +10,13 @@ class Job < ActiveRecord::Base
...
+      where(arel_table[:title].matches("%#{query}%").or(arel_table[:description].matches("%#{query}%")))
```
sophiedeziel: You can simplify that line with good ol' sql
```
where("title LIKE :query OR description LIKE :query", query: "%#{query}%")
```
(note that I may have made a typo or a syntax mistake)

bellef:
Thanks @sophiedeziel for finishing this! 😅

Issue #250, opened 2017-03-03
Error in sending reset-password instructions


marina101:
If you enter your email under "reset password instructions" (and the email has already been registered), when you click "Send me reset password instructions", the window displays "something went wrong".

sophiedeziel:
Thank you Marina for reporting it!

I think it's simply that SMTP is not set on our prod server. 

Talk to me on Slack if you really need to reset your password so I can help you.

sophiedeziel:
In progress. Waiting for verification on Mailgun's side.

sophiedeziel:
Yay! It's fixed! (but the name is 'please-fixme-in-config'

<img width="631" alt="screen shot 2017-03-28 at 2 34 10 pm" src="https://cloud.githubusercontent.com/assets/1037105/24421313/ca176dc8-13c3-11e7-890b-44f9ee4bbca6.png">

bellef:
I tried something [here](https://github.com/montrealrb/Montreal.rb/pull/269), let me know if it's ok for you @sophiedeziel

PR #269, opened 2017-05-15
Configure email sender (see #250) to 'courrier@sophiedeziel.com'

bellef:
Fix for https://github.com/montrealrb/Montreal.rb/issues/250
@sophiedeziel is it ok to put your email here?


sophiedeziel:
I think you should reopen that pull-request, but using an environment variable instead of a string. That way, it's a matter of configuration on the prod server.

Issue #261, opened 2017-03-24
reorder some resources in administrate


sophiedeziel:
In administrate, on the index of resources, they should be ordered in a logic manner:

for news and events, theys should be ordered by date (desc)
for locations and organization by name, alphabetically.
Jobs should have `published_at` displayed and ordered by it.

The order of the other pages does not bother me for now. If you can find a useful manner, i'll take it!

PR #272, opened 2017-05-24
[Feature][Issue 261] Populate job's published_at field

dechuck:
✅  ⚠️ Migration #271 ⚠️  ✅ 

One of the request in #261 is to reorder the jobs by the published at date.
However, this field isn't available on the jobs' table.

This PR will set the published_at field when a job is published.

It also contains a rake task to populate the published_at date on the older jobs.

Sorry for these 3 PRs


spec/lib/tasks/jobs.rake
```diff
@@ -0,0 +1,47 @@
...
+    context "when there's no published_at date" do
+      let!(:jobs) { create_list(:job, 3, :published) }
+
+      before do
+        subject.invoke
+      end
+
+      it "populate the published_at date" do
+        expect(
+          jobs.all? { |job| job.published_at.present? }
```
sophiedeziel: You can use a form like `expect(jobs.pluck(:published_at)).to all be_present`

The `all` helper is very useful for collections

spec/lib/tasks/jobs.rake
```diff
@@ -0,0 +1,47 @@
...
+        ).to be_truthy
```
sophiedeziel: I try to avoid `be_truthy` as much as possible. `eq true` will not lead to false positive in the case the return value is not a boolean.

app/models/job.rb
```diff
@@ -4,6 +4,8 @@ class Job < ActiveRecord::Base
   include Authorable
   STATES = %w(draft published archived).freeze
 
+  before_save :set_published_at, if: proc { state_changed?(to: "published") }
```
sophiedeziel: Fancy!

spec/lib/tasks/jobs.rake
```diff
@@ -0,0 +1,47 @@
...
+      end
+    end
+  end
+
+  context "When the jobs aren't published" do
+    let!(:jobs) { create_list(:job, 3, :draft) }
+
+    before do
+      subject.invoke
```
sophiedeziel: I think you can use only one `before` block for all the contexts since they are the same.

spec/support/shared_context/rake_helper.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+# https://robots.thoughtbot.com/test-rake-tasks-like-a-boss
```
sophiedeziel: 🤘

spec/support/shared_context/rake_helper.rb
```diff
@@ -0,0 +1,23 @@
...
+require "rake"
+
+shared_context "rake" do
```
sophiedeziel: You should exclude that file (or it's directory) from SimpleCov so it's not included in coverage stats.

app/models/job.rb
```diff
@@ -20,4 +22,10 @@ class Job < ActiveRecord::Base
             presence: true,
             inclusion: { in: STATES }
   validates :author, presence: true
+
+  private
+
+  def set_published_at
+    self.published_at = DateTime.current
```
dechuck: I'll find how `.touch` save the date in the Database.

However, since touch is call in the background when a record is saved, we should be good, since it will be in the same timezone.

lib/tasks/jobs.rake
```diff
@@ -0,0 +1,7 @@
+namespace :jobs do
+  desc 'Populate published_at field'
+  task populate_job_field: :environment do
+    Job.where(published_at: nil, state: :published)
```
dechuck: Happy you also like this approach. I'm prefer to have the data manipulation method inside rake task instead of migration. I'm always nervous with DB 😛

spec/lib/tasks/jobs.rake
dechuck: I didn't know this helper, thanks!

spec/lib/tasks/jobs.rake
dechuck: Good point!

spec/support/shared_context/rake_helper.rb
dechuck: I will rename the file to `rake_helper_spec`, this way it shouldn't be consider by SimpleCov. If it doesn't change anything, I will exclude the file manually.

sophiedeziel: That rename seems like a clever idea, but it's lying on the content of the file. I think the ideal would be to filter out `spec/support/**.*` altogether.

app/models/job.rb
dechuck: `.touch` is using the method `. current_time_from_proper_timezone `(https://github.com/rails/rails/blob/33e60514aed85b3076f2636d5f1ccfb513aace1c/activerecord/lib/active_record/timestamp.rb#L77) to generate the timestamp. 

By default, the timezone is set to `UTC` and it wasn't change in the config.

spec/lib/tasks/jobs.rake
dechuck: I try it out, but it doesn't work because the invoke is called before the jobs are created 😞

dechuck:
@sophiedeziel Are we allowed to rebase our branch? I'd like to squashed the last 4 commits together.

sophiedeziel:
Yes, you can rebase and squash if you want to, but there is no real benefit from it because I use the Squash merge feature from Github when I accept a pull-request.

app/models/job.rb
```diff
@@ -19,4 +21,11 @@ class Job < ActiveRecord::Base
   validates :state,
             presence: true,
             inclusion: { in: STATES }
+
+  private
+
+  def set_published_at
+    # We can't use touch since it only work on persisted object
```
sophiedeziel: 👍

dechuck:
Good to know. I won't rebase it then 😄

PR #275, opened 2017-09-15
Update Rubocop gem and ruby-style

dechuck:
I updated rubocop to the last version since there's a security issue in the version 0.4.5.

I disabled some rules to make sure the current code is still valid.

However, it could be interesting to enable `Style/SymbolArray`, `Lint/UriEscapeUnescape` and `Layout/EmptyLineAfterMagicComment`. What do you think?


.ruby-style.yml
```diff
@@ -233,6 +233,9 @@ Style/SpecialGlobalVars:
   Description: Avoid Perl-style global variables.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#no-cryptic-perlisms
   Enabled: false
+Style/SymbolArray:
+  Description: Check for array literals made up of symbols that are not using the %i() syntax
+  Enabled: false
```
sophiedeziel: I like that rule, but that is a matter of personal preference. I don't mind disabling it. Maybe if we dig a bit deeper to see if there is a technical advantage to use the `%i()` syntax? Or not :P

.ruby-style.yml
```diff
@@ -242,6 +245,9 @@ Style/WhenThen:
   Description: Use when x then ... for one-line cases.
   StyleGuide: https://github.com/bbatsov/ruby-style-guide#one-line-cases
   Enabled: false
+Layout/EmptyLineAfterMagicComment:
+  Description: Checks for a newline after the final magic comment.
+  Enabled: false
```
sophiedeziel: We can enable it, because that rule can autofix with the `-a` option

.ruby-style.yml
dechuck: Personnaly, I really like literals. I think the only advantage is that the are shorter to write. 

The rubobop auto correct function should catch them all 🙂

.ruby-style.yml
dechuck: Perfect 🙂

app/models/user.rb
```diff
@@ -37,8 +38,7 @@ def from_omniauth(auth)
 
     def new_with_session(params, session)
       super.tap do |user|
-        if data = session["devise.github_data"] &&
-                  session["devise.github_data"]["extra"]["raw_info"]
+        if data = session["devise.github_data"]&.dig("extra", "raw_info")
```
steakunderscore: Correct me if I'm wrong, I haven't used dig much, but should

```
if data = session["devise.github_data"] && session["devise.github_data"]["extra"]["raw_info"]
```

should be replaced with:

```
if data = session.dig("devise.github_data", "extra", "raw_info")
```

dechuck: We can't in this case, since session isn't a hash but a `ActionDispatch::Request::Session`.

I need to retrieve the hash first with `session["devise.github_data"]`. Also, I'm using `&.`, since `session["devise.github_data"]` could be nil.

Guardfile
```diff
@@ -1,4 +1,5 @@
 # frozen_string_literal: true
+
```
sophiedeziel: Haha! I see you runned rubocop 😉 👍

sophiedeziel:
Since we're here, I think it would be a good idea for new contributors to see in the docs that `rubocop -a` exists and should be ran before opening a pull-request to help in the "making code-climate happy" process!

dechuck:
Good point! I'm adding a section in the README :)

PR #276, opened 2017-09-18
Let members submit job posting

### LINKED ISSUE #254: As a logged in user, I want to submit a job posting
The path to completion is:

- [ ] On the jobs page, there is a `submit job posting` button always visible
- [ ] When the user is not logged in, it should be prompted the sign in page and then be redirected to the form (automatic with devise and `:authenticate_user!`)
- [ ] Of course, an already logged in user will see directly the form
- [ ] Should be in a regular view/controller (no administrate)
- [ ] the job lint output (see the admin section) should be visible
- [ ] the id of the member (as opposed to user) should be saved
- [ ] when submitted, the post should be in a Draft state
- [ ] Confirmation page should show the job posting + a mention of its draft state
- [ ] There should be a form to create an organization with the post (we don't care about duplicates for now)

dechuck:
Members can now submit job posting directly through the webiste.
When they write the job posting, the joblinter validate if the
offer respond to all the criteria.

fix #254 

capture d ecran 2017-09-17 a 17 32 13
capture d ecran 2017-09-17 a 17 32 21
capture d ecran 2017-09-17 a 17 36 12
capture d ecran 2017-09-17 a 17 36 43


sophiedeziel:
Oh wow!!!! I'm super happy with the UI! I wanted that feature for a long time 😲 

I'll review the code very soon!

dechuck:
Fantastic, I'm happy you like it 😄

app/assets/javascripts/job.coffee
```diff
@@ -0,0 +1,48 @@
+$(document).ready ->
```
sophiedeziel: With turbolinks, this will only work on page reloads. Use instead:
```
$(document).on 'turbolinks:load', ->
```

app/controllers/jobs_controller.rb
```diff
@@ -2,6 +2,37 @@
...
+      flash[:notice] = "The job posting has been send for review. Thanks!"
+      redirect_to confirmation_job_url(@job)
+    else
+      render :new
+    end
+  end
+
+  # GET /jobs/1/confirmation
+  def confirmation
+    @job = Job.draft.where(user_id: current_user&.id).find_by_id(params[:id])
```
sophiedeziel: This is nice scoping, but you can also use Pundit for authorizations.
```
@job = Job.find params[:id]
authorize(@job)
...
```

You'll have to define the `JobPolicy` class if it's not already defined and define the `confirmation?` method. You should also authorise in `new` and `create`.

dechuck: I never used Pundit. It seems really interesting! I'll take a look to the doc. Thanks for the heads up :)

sophiedeziel: I've used it for years! If you want to pair program a bit or see some live examples, I can find some time!

dechuck:
@sophiedeziel Let me know if you should change something else :)

app/controllers/jobs_controller.rb
```diff
@@ -2,6 +2,43 @@
...
+    else
+      render :new
+    end
+  end
+
+  # GET /jobs/1/confirmation
+  def confirmation
+    # Pundit need a validate Job object
+    # if find_by_id return nil, we set an empty Job
+    @job = Job.find_by_id(params[:id]) || Job.new
```
sophiedeziel: You can simply do `Job.find params[:id]` without the or, which will raise a 404 if not found. The authorize will to the rest if one if found

dechuck:
Should be good now :)

Issue #277, opened 2017-09-21
Feature: Embed youtube videos in talks


sophiedeziel:
The description is self-explanatory!
On the event view, when the talk is in the past, display the youtube video embeded. 

Maybe the description should be shortened with a "Read more" (all front-end, with a slide down animation)

opiation:
This seems like potentially two different features that may be independent of each other.  Perhaps the truncated _preview_ description could be tracked in a separate issue.

Regarding the YouTube vids, where are the videos currently stored/displayed?  In looking at the past year or so of events, on the **events** page, I did not find any YouTube links.  Have I missed something obvious or are they not displayed anywhere and simply stored in a DB or something?

sophiedeziel:
Haha not so self-explanatory after all.

Videos are on youtube and the links are not stored in the database yet. We only started recording them in September and shared the links through slack, facebook and twitter.

It should (obviously) only be visible on past events. 

The "Read More" could be a separate issue, let's display the thing first.

opiation:
As far as I can tell, there are 2 videos available (at least on the [**Montreal.rb** channel][0]):

- [Roberta Voulon](https://www.youtube.com/watch?v=GSmwG2aZwEI)
- [Anna Filina](https://www.youtube.com/watch?v=s0gKePTF31U)

Is this correct?

[0]: https://www.youtube.com/channel/UCmq5nOaMp18rcljeFocUUqw/videos

sophiedeziel:
Yes, and three others are coming.

opiation:
Thanks for your feedback.  I propose the following:

- Add a new `youtube_id` field to the `Talk` model to hold YouTube video ID hash (e.g: **GSmwG2aZwEI**)
- Add `youtube_id` input in related admin form
- When listing past events/talks, YouTube players are embedded between the talk title and the talk level (effectively placing the level below the video)
- The video pane should be approximately the width of the column containing the talk description

Does this sound about right?

sophiedeziel:
Yes! I'll review when I find a few minutes. Thank you for your pull-request!

PR #281, opened 2018-02-09
Added YouTube video ID to talks and embedded videos for past talks

opiation:
The pull request introduces embedded YouTube videos for past talks as proposed in #277.

This feature does the following:

- Adds a `youtube_id` string column to the **Talk** model
- Adds the `youtube_id` field to the talk dashboard
- Adds the "YouTube ID" label to the I18n config/locales/en.yml
- Adds an embedded `<iframe>` video if the talk is both in the past **and** has a non-nil `youtube_id`


opiation:
Should I presume from the above coverage decrease alert that that PR will not be accepted on those grounds?

sophiedeziel:
Yep. I can do the review in parallel, but coverage requirement has to be met.

opiation:
No problem.  I'll add some coverage tonight.

app/views/talks/_talk.html.erb
```diff
@@ -9,6 +9,11 @@
 
       <h4><%= render_markdown_as_html "#{talk.title.titleize}, by #{talk.member.name}" %></h4>
 
+      <% if talk.event&.starts_at&.past? && talk.youtube_id.present? %>
+        <iframe src="https://www.youtube.com/embed/<%= talk.youtube_id %>?rel=0&autoplay=0" frameborder="0" allowfullscreen width="640" height="360"></iframe>
```
sophiedeziel: This pull-request is pretty simple and very good. The only suggestion I would make is create a helper for the youtube tag

sophiedeziel:
I'm merging, even with my suggestion. I understand that you might not have the time to make that change and this is adding a lot of value to the project as is. Thank you for opening that pull-request!

PR #288, opened 2018-03-23
%s/FactoryGirl/FactoryBot/g

lucianosousa:
see https://github.com/thoughtbot/factory_bot/blob/master/NAME.md


spec/spec_helper.rb
```diff
@@ -16,7 +16,7 @@
 ActiveRecord::Migration.check_pending! if defined?(ActiveRecord::Migration)
 
 RSpec.configure do |config|
-  config.include FactoryGirl::Syntax::Methods
+  config.include FactoryBot::Syntax::Methods
```
denispasin: Do we need this twice ? cf: ` spec/support/factory_girl.rb` also should rename the support file.

db/seeds.rb
```diff
@@ -2,17 +2,17 @@
 
 require "factory_girl_rails"
```
denispasin: Sorry, I missed that one...

lucianosousa: @denispasin done

lucianosousa:
🎆

PR #295, opened 2018-06-25
Add Category for Today's Event

peacestone:
When a user views the home page or events index page, if there is an event today the user will see it in its own section.


peacestone:
Looks like I forgot to update the spec. I'll have to figure out how to update the tests. Any help would be appreciated.

